PR #1, opened 2020-03-11
Adds the biaffine dependency parser

dirkgr:
I'm doing this as a PR to test-drive the process of creating PRs against this repo.


dirkgr:
I'm seeing some hard-to-believe numbers from the timings with that pip dependency cache. I think I'll leave it in for now and see how it goes, if only to not hammer the pip repositories.

.github/workflows/main.yml
```diff
@@ -15,21 +15,27 @@ jobs:
...
+      with:
+        path: ~/.cache/pip
+        key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
+        restore-keys: |
+          ${{ runner.os }}-pip-
     - name: Install requirements
       run: |
         pip install -r requirements.txt
         pip install pytest pytest-cov flake8 black mypy
-    - name: Run tests
```
schmmd: Why remove tests?  I don't know much about the GitHub actions as Mark set it up.

matt-gardner: It was just moved below. I curious to know why you moved it, though.

dirkgr: Because linting is fast and tests are slow. I don't want to wait for slow tests just to have the 5s linting test fail afterwards.

PR #12, opened 2020-04-04
MST decoding

dirkgr:
Having MST decoding off breaks when it runs as a predictor, in the demo. It would be easiest if we could just train with it on. I'm trying that now and comparing the scores.

### DIFF
--- a/allennlp_models/syntax/biaffine_dependency_parser/biaffine_dependency_parser.py
+++ b/allennlp_models/syntax/biaffine_dependency_parser/biaffine_dependency_parser.py
@@ -593,7 +593,10 @@ class BiaffineDependencyParser(Model):
             instance_head_tags[0] = 0
             heads.append(instance_heads)
             head_tags.append(instance_head_tags)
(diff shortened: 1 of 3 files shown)


dirkgr:
Scores are virtually identical. Before:
```
  "best_validation_UAS": 0.9500837988826816,
  "best_validation_LAS": 0.9381284916201117,
  "best_validation_UEM": 0.55,
  "best_validation_LEM": 0.48705882352941177,
  "best_validation_loss": 0.464306292789323
```
After:
```
  "best_validation_UAS": 0.9500558659217877,
  "best_validation_LAS": 0.9381284916201117,
  "best_validation_UEM": 0.5547058823529412,
  "best_validation_LEM": 0.49117647058823527,
  "best_validation_loss": 0.464306292789323
```

matt-gardner:
Why does it break when running in the demo?  And does this affect running times?

dirkgr:
It breaks in the demo because only with MST are we guaranteed a well-formed tree.

dirkgr:
There is no practical difference in training times. I'm not sure that algorithm is slow anyways, because the sentences are pretty short. Also, the change only applies to validation, and the validation set is not that big.

PR #16, opened 2020-04-08
Adds NLI configs

dirkgr:
Also moves the coref predictor, which I forgot earlier.

See https://github.com/allenai/allennlp/pull/4043.


dirkgr:
As usual, this won't work until  allenai/allennlp#4043 is merged and this PR points to the new version of the core library.

PR #20, opened 2020-04-17
Pretty up README.md

epwalsh:


epwalsh:
> Why did you decide to make this HTML instead of Markdown?

Just so that the image size can be tweaked and the content can be centered. I just thought it looked a little nicer that way

PR #22, opened 2020-04-20
ensure setuptools and wheel installed first

epwalsh:


epwalsh:
> I'm assuming you found a problem caused by not having these?

yea in some cases the build for other packages would fail because wheel hadn't been installed yet

PR #30, opened 2020-04-27
Add more checks to release workflow

epwalsh:
There's a lot going on here, but essentially we're just bringing the models CI up to speed with all of the improvements we've made to the core repo CI. This also adds a check to make sure the models plugin is discoverable.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -114,6 +114,10 @@ jobs:
       run: |
         pip install $(ls dist/*.whl)
 
+    - name: Ensure models plugin found
+      run: |
+        python -c 'from allennlp.common.plugins import discover_plugins; assert "allennlp_plugins.allennlp_models" in list(discover_plugins())'
+
   publish:
     name: PyPI
     needs: [check, build_package, test_package]
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,8 +5,8 @@ on:
     types: [published]
 
 jobs:
-  check:
-    name: Check
+  build:
+    name: Build
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -40,6 +56,68 @@ jobs:
       run: |
         python setup.py bdist_wheel sdist
 
+    - name: Save package
+      if: matrix.python == '3.7'
```
dirkgr, line 60: Can we upload a version for both, since we're already here?

.github/workflows/release.yml
```diff
@@ -40,6 +56,68 @@ jobs:
...
+        name: models-package
+        path: dist
+
+    - name: Install models package
+      run: |
+        pip install $(ls dist/*.whl)
+
+    - name: Ensure models plugin found
+      run: |
+        python -c 'from allennlp.common.plugins import discover_plugins; assert "allennlp_plugins.allennlp_models" in list(discover_plugins())'
```
dirkgr, line 96: 👍

scripts/get_version.py
```diff
@@ -0,0 +1,45 @@
...
+def parse_args():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("version_type", choices=["stable", "latest", "current"])
+    return parser.parse_args()
+
+
+def get_current_version() -> str:
+    VERSION: Dict[str, str] = {}
+    with open("allennlp_models/version.py", "r") as version_file:
+        exec(version_file.read(), VERSION)
```
dirkgr, line 18: Couldn't you just import that file?

.github/workflows/release.yml
dirkgr: I love the thoroughness of this. Something went wrong, so here is some code to make sure next time we get an error first. Awesome!

scripts/get_version.py
epwalsh: Well yes hahah, I copied this from the core repo without thinking much. But actually I do think it's beneficial to keep it like this because
(A) it's faster than importing
(B) in some cases we might want to run this script without actually installing `allennlp_models` or any of its dependencies

.github/workflows/release.yml
epwalsh: Well we only end up pushing one build to PyPI (arbitrarily the one built in 3.7), but that distribution has to work for both 3.6 and 3.7. So I figured I would structure the CI in the same way: build the package in one version of Python but test it across all versions.

PR #32, opened 2020-04-29
add install instructions to README

epwalsh:
This should clarify https://github.com/allenai/allennlp/issues/4155

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,4 +56,4 @@ EXCLUDE_ALLENNLP_IN_SETUP=true pip install -e .
 pip install -r dev-requirements.txt
 ```
 
-Both `allennlp` and `allennlp-models` are developed and tested side-by-side, so they should be kept up-to-date with each other. If you look at the GitHub Actions [workflow for allennlp-models](https://github.com/allenai/allennlp-models/actions), it's always tested against the master branch of allennlp. Similarly, allennlp is always tested against the master branch of allennlp-models.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -22,4 +22,36 @@
...
+pip install allennlp-models==1.0.0rc3
+pip freeze | grep allennlp
+# > allennlp==1.0.0rc3
+# > allennlp-models==1.0.0rc3
+```
+
+### From source
+
+If you intend to install the models package from source, then you probably also want to [install `allennlp` from source](https://github.com/allenai/allennlp#installing-from-source).
+Once you have `allennlp` installed, run the following within the same Python environment:
```
nelson-liu, line 51: what version of allennlp should you be using with the latest models version? is it expected that models is updated lockstep with allennlp-core (i.e., always make sure you have the latest master for both)?

epwalsh: They are developed and tested side-by-side, so they should be kept up-to-date with each other. If you look at the GitHub Actions workflow for `allennlp-models`, it's always tested against the master branch of `allennlp`. Similarly, `allennlp` is always tested against the master branch of `allennlp-models`.

nelson-liu: That's good to know, thanks! Do you think that information should be added to the PR? I guess i'm imagining a scenario where I have to go to a previous commit of allennlp-model or allennlp-core, and your information would be helpful in figuring out what corresponding other hash i need.

epwalsh: Just added!

PR #33, opened 2020-04-29
Depend on an unreleased version of AllenNLP

dirkgr:
The latest daily isn't out yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,8 +52,10 @@ Once you have `allennlp` installed, run the following within the same Python env
 
 ```
 git clone https://github.com/allenai/allennlp-models.git
-EXCLUDE_ALLENNLP_IN_SETUP=true pip install -e .
+ALLENNLP_VERSION_OVERRIDE='allennlp' pip install -e .
 pip install -r dev-requirements.txt
 ```
+
+The `ALLENNLP_VERSION_OVERRIDE` flag ensures that your local install of `allennlp` won't be overridden. Otherwise `allennlp` will be intalled from the latest commit to the master branch on GitHub.
 
 Both `allennlp` and `allennlp-models` are developed and tested side-by-side, so they should be kept up-to-date with each other. If you look at the GitHub Actions [workflow for `allennlp-models`](https://github.com/allenai/allennlp-models/actions), it's always tested against the master branch of `allennlp`. Similarly, `allennlp` is always tested against the master branch of `allennlp-models`.
--- a/README.md
+++ b/README.md
@@ -52,10 +52,11 @@ Once you have `allennlp` installed, run the following within the same Python env
 
 ```
 git clone https://github.com/allenai/allennlp-models.git
+cd allennlp-models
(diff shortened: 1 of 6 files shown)


epwalsh:
I'm confused by why we need this, and it will break the logic in `setup.py` https://github.com/allenai/allennlp-models/blob/master/setup.py#L30

dirkgr:
Yeah, I know. I hope this fixes it. What a mess.

dirkgr:
The root of the problem is that `requirements.txt` and `setup.py` need a different format for depending on URLs, but we treat them the same.

The fact that I wanted to do this for the allennlp dependency is somewhat orthogonal.

dirkgr:
This changes the flag we use to override the allennlp dependency. I think it makes the code in `setup.py` more readable, but I don't know if it covers all the use cases we care about. @epwalsh?

epwalsh:
I guess my concern is that in the CI for this repo, we wouldn't be testing against the latest commit of allennlp, and we'd have to keep manually updating the pinned commit.

dirkgr:
Ah, yes. So we do want "exclude" behavior, not "override". But we also need a way to pin it to something other than the latest rc.

epwalsh:
Having exclude behavior may not be necessary as long as there is a way to have the allennlp dependency completely unpinned.

epwalsh:
Sorry for sneaking in a couple commits, just wanted to update the README before I forgot

README.md
```diff
@@ -52,8 +52,10 @@ Once you have `allennlp` installed, run the following within the same Python env
 
 ```
 git clone https://github.com/allenai/allennlp-models.git
-EXCLUDE_ALLENNLP_IN_SETUP=true pip install -e .
+ALLENNLP_VERSION_OVERRIDE='allennlp' pip install -e .
 pip install -r dev-requirements.txt
 ```
 
+The `ALLENNLP_VERSION_OVERRIDE` flag ensures that your local install of `allennlp` won't be overridden. Otherwise `allennlp` will be intalled from the latest commit to the master branch on GitHub.
```
nelson-liu, line 59: s/intalled/installed/

requirements.txt
```diff
@@ -1,4 +1,5 @@
-allennlp
+git+https://github.com/allenai/allennlp.git
```
nelson-liu, line 1: maybe explicitly specify `allennlp.git@master?`

README.md
```diff
@@ -52,8 +52,10 @@ Once you have `allennlp` installed, run the following within the same Python env
 
 ```
 git clone https://github.com/allenai/allennlp-models.git
-EXCLUDE_ALLENNLP_IN_SETUP=true pip install -e .
+ALLENNLP_VERSION_OVERRIDE='allennlp' pip install -e .
```
nelson-liu, line 55: sorry this is an unrelated change, but you probably also want to add a `cd allennlp-models` above this line?

requirements.txt
dirkgr: sure

PR #34, opened 2020-04-29
Plugin package should be namespace package

epwalsh:
I believe this is how the package was intended to be built. `allennlp_plugins` is not an actual package (because no `__init__.py`), but `allennlp_plugins.allennlp_models` is a namespace package.


setup.py
```diff
@@ -94,7 +94,17 @@ def fix_url_dependencies(req: str) -> str:
     author="Allen Institute for Artificial Intelligence",
     author_email="allennlp@allenai.org",
     license="Apache",
-    packages=find_packages(exclude=["*.tests", "*.tests.*", "tests.*", "tests"],),
+    packages=find_namespace_packages(
```
dirkgr, line 97: Can't we just include what we want, instead of excluding what we don't want?

Also, this will treat the regular package as a "namespace package" as well. Is that acceptable?

epwalsh: Using include led to that bug before, so I’m trying to avoid it since the behavior is not consistent across environments. And I don’t think the main package is treated any differently, at least from what I could tell through building and installing it this way

epwalsh:
Closing because we got rid of this mechanism

PR #35, opened 2020-04-29
Implemented BART

Tobias-Rohde:
- implemented BART using huggingface implementation
- Bart class is the main model that can be used for text generation
- BartEncoder is a Seq2SeqEncoder that is extracted from the BART model
without the embeddings
- The Bart encoder can be swapped out with any AllenNLP Seq2SeqEncoder,
provided that it's output dimension matches that of the Bart decoder
- added a small testing file (bart_test.py), which contains code for
testing the implementation. Before merging, this should be rewritten as
a proper unit test.


allennlp_models/seq2seq/bart/bart.py
```diff
@@ -0,0 +1,401 @@
...
+        input_ids, input_mask = inputs["tokens"]["token_ids"], inputs["tokens"]["mask"]
+
+        outputs = {}
+
+        # TODO: (not targets) => generating?
+        # assert targets is None or self.training  # Don't provide targets for generation
+        generating = not self.training
+
+        # If no targets are provided, then shift input to right by 1. Bart already does this interally
+        # but it does not
```
dirkgr, line 197: But it does not what?

allennlp_models/seq2seq/bart/bart.py
```diff
@@ -0,0 +1,401 @@
...
+            cache = decoder_cache[layer_idx]
+            if attention_name not in cache:
+                cache[attention_name] = {}
+            assert tensor_name not in cache[attention_name]
+            cache[attention_name][tensor_name] = cache_value
+        return decoder_cache
+
+    def take_step(
+        self, last_predictions: torch.Tensor, state: Dict[str, torch.Tensor], step: int
+    ) -> Tuple[torch.Tensor, Dict[str, torch.Tensor]]:
```
dirkgr, line 293: Do you have to turn off gradients somewhere in here, or is that taken care of outside?

allennlp_models/seq2seq/bart/bart.py
```diff
@@ -0,0 +1,401 @@
...
+            tokens.
+
+        """
+        predictions = output_dict["predictions"]
+        predicted_tokens = [None] * predictions.shape[0]
+        for i in range(predictions.shape[0]):
+            predicted_tokens[i] = self._indexer.indices_to_tokens(
+                {"token_ids": predictions[0].tolist()}, self.vocab
+            )
+        output_dict["predicted_tokens"] = predicted_tokens
```
dirkgr, line 391: Do these functions usually modify their inputs? If so, carry on. If not, I'd prefer it if this one also didn't.

allennlp_models/seq2seq/bart/bart_test.py
```diff
@@ -0,0 +1,82 @@
...
+from allennlp.nn.util import move_to_device
+
+import torch
+
+# from transformers import BartConfig
+
+from allennlp_models.seq2seq import Bart
+
+
+# TODO: remove this file, it's just for testing.
```
dirkgr, line 14: Then it should not be in the PR.

We like tests though! How much work is it to add a minimal test for BART?

allennlp_models/seq2seq/seq2seq_reader.py
```diff
@@ -109,23 +121,28 @@ def _read(self, file_path: str):
     def text_to_instance(
         self, source_string: str, target_string: str = None
     ) -> Instance:  # type: ignore
+        if self._ptb_sanitize_strings:
+            source_string = sanitize_ptb_tokenized_string(source_string)
+            target_string = sanitize_ptb_tokenized_string(target_string)
```
dirkgr, line 126: Since you found a better source file, this is no longer necessary, right?

dirkgr: To be sure, I want to keep the `sanitize_ptb_tokenized_string` function, but unless there is a compelling reason to add this functionality to the seq2seq reader, we don't have to make it in here.

allennlp_models/seq2seq/bart/bart.py
```diff
@@ -0,0 +1,401 @@
...
+    @overrides
+    def forward(self, inputs: torch.Tensor, mask: torch.BoolTensor):
+        # The first element is always the last encoder states for each input token.
+        # Depending on the config, the second output will contain a list of the encoder states
+        # after each transformer layer. Similarly, the third output can contain the attentions from each layer.
+        # We only care about the first element.
+        return self.bart_encoder(input_ids=inputs, attention_mask=mask)[0]
+
+
+class _BartEncoderWrapper(nn.Module):
```
dirkgr, line 58: So `_BartEncoderWrapper` lets you use any AllenNLP seq2seq module as encoder, and `BartEncoder` lets you plug the original BART encoder back in? `b = Bart(..., encoder=None)` functions the same as `b = Bart(..., encoder=BartEncoder())`?

dirkgr:
Also, can you point this at your branch of the core library, so we can see the tests pass?

allennlp_models/seq2seq/bart/bart.py
```diff
@@ -0,0 +1,401 @@
...
+from torch import nn
+
+
+@Seq2SeqEncoder.register("bart_encoder")
+class BartEncoder(Seq2SeqEncoder):
+    """
+    The BART encoder without the token and position embeddings.
+    """
+
+    def __init__(self, model_name):
```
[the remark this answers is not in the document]
Tobias-Rohde: Done.

allennlp_models/seq2seq/bart/bart.py
Tobias-Rohde: It doesn't use those for the loss calculation, I fixed the comment.

allennlp_models/seq2seq/bart/bart.py
Tobias-Rohde: Good catch, this doesn't seem to be done anywhere, also not for the other seq2seq classes that use our beam search code. I added a `torch.no_grad` to the beam search code so that we disable it in all of those cases. I'll push this to the bart branch on allennlp.

allennlp_models/seq2seq/bart/bart.py
Tobias-Rohde: Yes, in fact this is explicitly noted in the comment of the base method:

`This method `modifies` the input dictionary, and also `returns` the same dictionary.`

allennlp_models/seq2seq/bart/bart_test.py
Tobias-Rohde: I removed the file, I'll make a test once I am sure everything is working correctly.

allennlp_models/seq2seq/seq2seq_reader.py
Tobias-Rohde: Yes, I wrote my own preprocessing code for the CNN/DM data, so I don't have to bother with the PTB tokenizer anymore. I removed those calls.

allennlp_models/seq2seq/bart/bart.py
Tobias-Rohde: Yes, that would work, but is clearly useless. The idea behind BartEncoder is that you can use it as part of a different model.

Tobias-Rohde:
> Also, can you point this at your branch of the core library, so we can see the tests pass?

I am not exactly sure how to do that. I changed the branch in `requirements.txt`, is that they way to do it?

dirkgr:
Does that training config work, right now?

allennlp_models/seq2seq/bart/bart.py
```diff
@@ -199,6 +199,8 @@ def forward(
         # assert targets is None or self.training  # Don't provide targets for generation
         generating = not self.training
 
+        # assert not generating or self.bart.config.output_past
```
dirkgr, line 202: Debug code?

training_config/seq2seq/bart_cnn_dm.jsonnet
```diff
@@ -0,0 +1,52 @@
...
+        "source_token_indexers": {
+            "tokens": {
+                "type": "pretrained_transformer",
+                "model_name": "bart-large",
+                "namespace": "tokens"
+            }
+        },
+
+        "start_symbol": "<s>",
+        "end_symbol": "</s>",
```
dirkgr, line 20: Is that needed? Isn't that given by the transformer, i.e., we could get the right values from the huggingface tokenizer?

Tobias-Rohde: As discussed, this is something we can only get form the PretrainedTransformerTokenzier, but in the Seq2SeqReader we only have access to the AllenNLP tokenizer which doesn't have start/end symbol properties (yet).

allennlp_models/seq2seq/bart/bart.py
Tobias-Rohde: Removed!

training_config/seq2seq/bart_cnn_dm.jsonnet
```diff
@@ -0,0 +1,58 @@
+{
+    "train_data_path": "",
+    "validation_data_path": "",
+    "dataset_reader": {
+        "type": "allennlp_models.generation.dataset_readers.Seq2SeqDatasetReader",
```
dirkgr, line 5: Do you have to specify the whole name?

training_config/seq2seq/bart_cnn_dm.jsonnet
```diff
@@ -0,0 +1,56 @@
+local model_name = "facebook/bart-large";
+
+{
+    "train_data_path": "/home/tobiasr/Documents/AllenNLP/data/train.csv",
```
dirkgr, line 4: This data is public, right? So we could upload it to our data bucket and refer to it there?

training_config/seq2seq/bart_cnn_dm.jsonnet
```diff
@@ -0,0 +1,56 @@
...
+        "optimizer": {
+            "type": "huggingface_adamw",
+            "lr": 3e-5,
+            "betas": [0.9, 0.999],
+            "eps": 1e-8,
+            "correct_bias": true
+        },
+        "learning_rate_scheduler": {
+            "type": "polynomial_decay",
+            "total_steps": 215334
```
dirkgr, line 47: I don't think this is necessary anymore. It should figure out the number of steps automatically now.

dirkgr:
Are there still issues with pre-processing the data?

dirkgr:
Note to self: The dataset reader reads the input data as presented by https://drive.google.com/uc?export=download&id=0BwmD_VLjROrfTHk4NFg2SndKcjQ and https://drive.google.com/uc?export=download&id=0BwmD_VLjROrfM1BxdkxVaTY2bWs in two directories called `dailymail_stories` and `cnn_stories`. It also needs https://github.com/abisee/cnn-dailymail/tree/master/url_lists in a sub-folder called `url_lists`. Dirk will add some infrastructure so that the `cached_path()` mechanism can download directories.

allennlp_models/generation/dataset_readers/cnn_dm.py
```diff
@@ -0,0 +1,165 @@
...
+    source_token_indexers : `Dict[str, TokenIndexer]`, optional
+        Indexers used to define input (source side) token representations. Defaults to
+        `{"tokens": SingleIdTokenIndexer()}`.
+    target_token_indexers : `Dict[str, TokenIndexer]`, optional
+        Indexers used to define output (target side) token representations. Defaults to
+        `source_token_indexers`.
+    source_add_start_token : bool, (optional, default=True)
+        Whether or not to add `START_SYMBOL` to the beginning of the source sequence.
+    source_add_end_token : bool, (optional, default=True)
+        Whether or not to add `END_SYMBOL` to the end of the source sequence.
```
dirkgr: These are old. You already know that.

allennlp_models/generation/dataset_readers/cnn_dm.py
```diff
@@ -0,0 +1,165 @@
...
+
+    @staticmethod
+    def _hashhex(url):
+        h = hashlib.sha1()
+        h.update(url)
+        return h.hexdigest()
+
+    @staticmethod
+    def _sanitize_story_line(line):
+        line = ftfy.fix_encoding(line)
```
dirkgr, line 76: I think there are more aggressive functions in `ftfy` you could try here, but they all make it slower.

dirkgr:
I'll merge this and train a model for the model zoo later.

PR #42, opened 2020-05-04
Adds the `pretrained` module from allennlp-hub.

dirkgr:
These tests might not succeed because of the disk space issue.

Finishes https://github.com/allenai/allennlp/issues/3302.
Closes https://github.com/allenai/allennlp/issues/3965.

### DIFF
--- a/allennlp_models/pretrained.py
+++ b/allennlp_models/pretrained.py
@@ -1,5 +1,4 @@
 from allennlp.models import load_archive
-from allennlp.modules import conditional_random_field
 from allennlp.predictors import Predictor, SentenceTaggerPredictor
 
 from allennlp_models.coref import CorefPredictor
--- a/allennlp_models/pretrained.py
+++ b/allennlp_models/pretrained.py
@@ -1,0 +1,152 @@
+from typing import Type
+
+from allennlp.models import load_archive
+from allennlp.predictors import Predictor, SentenceTaggerPredictor
+
+from allennlp_models.coref import CorefPredictor
+from allennlp_models.nli import DecomposableAttentionPredictor
+from allennlp_models.rc.bidaf import ReadingComprehensionPredictor
+from allennlp_models.syntax import SemanticRoleLabelerPredictor, OpenIePredictor, ConstituencyParserPredictor, \
+    BiaffineDependencyParserPredictor
+
+
+def _load_predictor(archive_file: str, predictor_name: str) -> Predictor:
+    """
+    Helper to load the desired predictor from the given archive.
+    """
+    archive = load_archive(archive_file)
+    return Predictor.from_archive(archive, predictor_name)
+
+
+def bert_srl_shi_2019() -> SemanticRoleLabelerPredictor:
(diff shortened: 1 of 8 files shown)


dirkgr:
Our organization of pre-trained models is still a total mess. We have three places now where we have the URLs. Here, and twice in allennlp-models. They don't all agree, and they don't all have the same models. We will have to fix this soon. It's confusing.

epwalsh:
Looks like the tests you wrote aren't found by pytest. I think by default pytest assumes test classes are named `Test*` instead of `*Test`. But if we just add [this line](https://github.com/allenai/allennlp/blob/master/pytest.ini#L3) to the `pytest.ini` file it will find them.

epwalsh:
Another thing to consider is that we'll be re-downloading these models on each CI run, which will run up our cloud bill. If we used a self-hosted runner for the pretrained tests we wouldn't have this problem since the models would be cached locally.

allennlp_models/pretrained.py
```diff
@@ -0,0 +1,146 @@
...
+    Coreference Resolution
+    """
+    predictor = _load_predictor(
+        "https://storage.googleapis.com/allennlp-public-models/coref-spanbert-large-2020.02.27.tar.gz",
+        "coreference-resolution",
+    )
+    return predictor
+
+
+def named_entity_recognition_with_elmo_peters_2018() -> SentenceTaggerPredictor:
```
epwalsh, line 86: I think we'll need to import the crf tagger module for this to work

tests/pretrained_test.py
```diff
@@ -6,6 +8,10 @@
 from allennlp_models import pretrained
 
 
+# But default we don't run these tests
+@pytest.mark.skipif(
+    not os.environ.get("ALLENNLP_MODELS_RUN_PRETRAINED_TEST"), reason="requires massive downloads"
+)
```
dirkgr, line 14: The GPU tests are selected in a different way, aren't they? Why isn't this done the same way?

epwalsh: Oh yea, we should probably run those on the self-hosted runner as well

epwalsh: Sorry - I realize your question now. I felt like there be an explicit skip condition here because if you were running tests locally you might not want to download these massive files, and I didn't know what that condition could be other than an environment variable flag

epwalsh: But we could also add a custom mark like we do for GPU tests and then exclude those by default. That might be cleaner

dirkgr: Looks like we're inconsistent about how we handle GPU tests. Some have `@pytest.mark.gpu`, others use `skipif` and detect how many GPUs there are. Others yet have both.

epwalsh: The `skipif` ensures they're never accidentally run if you don't have a GPU available. The `mark.gpu` is just so that we can select only GPU tests to run

dirkgr:
I think this is good to go then?

PR #44, opened 2020-05-06
Be compatible with the new tokenizers code

dirkgr:
See https://github.com/allenai/allennlp/pull/3868


dirkgr:
The meaning of the token type ids for TransformerQA flips in this PR. I confirmed that I can still get a good score with it.

dirkgr:
I cannot think of a way in which all of these tests could pass, but the one model doesn't.

dirkgr:
Pretty sure this is the problem: https://github.com/allenai/allennlp-models/pull/44/checks?check_run_id=668818627#step:4:42

It used the cache for the `pip install -r requirements.txt` step, because `requirements.txt` didn't change. It still points at the latest master. But the latest master has changed in the meantime.

PR #45, opened 2020-05-07
update for new AllenNlpTestCase

epwalsh:
See https://github.com/allenai/allennlp/pull/4204


requirements.txt
```diff
@@ -1,4 +1,4 @@
-git+https://github.com/allenai/allennlp.git@master
+git+https://github.com/epwalsh/allennlp.git@test-case
```
epwalsh, line 1: Temporary change to make sure tests will pass

PR #51, opened 2020-05-13
Move seq2seq tests to right test directory

epwalsh:


epwalsh:
Oh, just realized the seq2seq tests were not missing, just in the wrong directory.

PR #54, opened 2020-05-13
add commit and release Dockerfiles with instructions

epwalsh:
Closes https://github.com/allenai/allennlp/issues/4222

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,13 +68,14 @@ whether you will leverage a GPU or just run on a CPU.  Docker provides more
 isolation and consistency, and also makes it easy to distribute your
 environment to a compute cluster.
 
-Once you have [installed Docker](https://docs.docker.com/engine/installation/)
-you can build an image with a specific version of `allennlp` and `allennlp-models` with
+Once you have [installed Docker](https://docs.docker.com/engine/installation/) you can either use a [prebuilt image from a release](https://hub.docker.com/r/allennlp/allennlp-models) or build an image locally with any version of `allennlp` and `allennlp-models`.
+
+To build an image locally from a specific release, run
 
    ```bash
    docker build \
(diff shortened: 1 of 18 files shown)


README.md
```diff
@@ -60,3 +60,41 @@ pip install -r dev-requirements.txt
...
+Docker provides a virtual machine with everything set up to run AllenNLP--
+whether you will leverage a GPU or just run on a CPU.  Docker provides more
+isolation and consistency, and also makes it easy to distribute your
+environment to a compute cluster.
+
+Once you have [installed Docker](https://docs.docker.com/engine/installation/)
+you can build an image with a specific version of `allennlp` and `allennlp-models` with
+
+   ```bash
+   docker build \
```
schmmd, line 75: I would use `docker build -f` instead of `<`.

README.md
```diff
@@ -60,3 +60,41 @@ pip install -r dev-requirements.txt
...
+       --build-arg ALLENNLP_VERSION=1.0.0rc3 \
+       -t allennlp-models - < Dockerfile.release
+   ```
+
+Just replace "1.0.0rc3" with the desired version.
+
+Alternately, you can build against specific commits of `allennlp` and `allennlp-models` with
+
+   ```bash
+   docker build \
```
schmmd, line 85: Same here.  Prefer `-f`.

README.md
epwalsh: I only used `<` since the image doesn't actually require a context

schmmd: Once you make this adjustment try this command verbatim (copy paste).  You might be missing a dot `.`

epwalsh: Hmm I actually think it's more clear to use `<`, i.e. to explicitly not use a context, so that it's obvious that we're not building the image from the local source code.

PR #58, opened 2020-05-22
Renames the model categories

dirkgr:
Fixes https://github.com/allenai/allennlp/issues/4051.

In effect, this just renames all the categories. They were already grouped in the way Matt wanted in https://github.com/allenai/allennlp/issues/4051, just the names were different.

The only exception is language modeling. I left that in its own category. I think it's weird to have that in a subdirectory of generation, but if you would prefer it there, I can move it.

I also left the internal structure of the directories alone, but if it makes you happy, I can flatten them all out.


nelson-liu:
Thanks for working on this! I still wish there was a bit more consistency in how the per-format directories are organized. For a module like RC, i think we should be using the same organization as the `allennlp` module (e.g., separate out `dataset_readers`, `models`, `metrics`).

Personally, I'd want that organization for _all_ the per-format modules. I can understand if people think the overhead is too much at the moment (e.g., we have a folder with that contains only a single dataset), but I'd argue that if you want to be adding more models to this repo over time, it's better to go with a consistent structure per-module so you can add to it (versus having to create a folder for things down the line and worrying about deprecations).

I must say that, personally speaking, I've avoided upgrading to 1.0 / latest master because I like having the monorepo with all of the models—I know exactly where things are, because it's organized very consistently. I guess I'd like to see that come back for this repo as well.

matt-gardner:
I was going to say the same thing as @nelson-liu.  As a concrete example, just a couple of days ago I went looking for the DROP evaluation script.  I did not know where to find it.  Turns out it's under `rc.qanet`, but I never would have guessed that, and I had to search all of the subdirectories.  I probably would have thought it wasn't in there and stopped looking, if I didn't know for sure it was there somewhere.  It's the organization of the subdirectories that's the major discoverability issue.

dirkgr:
You guys are not even remotely thinking of this the same way I am. I am imagining a world where we have a core library, and a bunch of models, each of which could stand on its own. Some models are first-party and we maintain them, and some others are third party, and they are maintained by someone else. We stick the models into a single repo because it's less overhead this way.

For your perspective, I surmise we have a core library with some tools, and then another library with some other tools?

I have most of the changes done by now, but this is taking us far away from having a real model library, like Hugginface or paperswithcode.com. What we have here is the opposite, and I would expect 90% of our users to just import both allennlp and allennlp-models and not think about the difference at all.

If these models are so core to the experience, why are they not in the core library?

nelson-liu:
(this is mostly just my opinion, I wasn't at AI2 when the bulk of the repo-splitting work commenced, so I might be missing some context here. Maybe just take it as opinions from someone who's used the library a lot?)

> If these models are so core to the experience, why are they not in the core library?

IIRC it was because the core library had too many dependencies / was too large. The original idea was to split things off into separate modules, and enable people to pick-and-choose what components they wanted. Personally, I'm fine with how long the package takes to download and install, but it's worth being mindful and inclusive of people in lower-resourced environments that might find it very painful.

> I have most of the changes done by now, but this is taking us far away from having a real model library, like Hugginface or paperswithcode.com. What we have here is the opposite, and I would expect 90% of our users to just import both allennlp and allennlp-models and not think about the difference at all.

I guess the way I think that this repo differs from huggingface transformers is that, in transformers, when you import the module, you're probably going to be training a model _from their repo_. I'm not sure that's the case here. Personally, running models on new data isn't that big of a use case for me. I'd guess that more researchers are _developing_ models, whereas in huggingface it's mostly taking something pretrained off the shelf and running it on new data. Furthermore, the models here are a lot more diverse than huggingface, because they have different input APIs and are only compatible with certain Instance formats—you cant switch and swap them like in huggingface.

Personally, when I use AllenNLP, the goal is to _build_ a novel model (rather than simply train one). My workflow looks something like this. Say I'm working on DROP. I would start a new file in my repo (I always use allennlp as a library, and never work directly on top of the repo). Then, I got to go find all of the imports I need—if things are consistent, I can just do `allennlp_models.rc.dataset_readers.drop`, `allennlp_models.rc.metrics.drop`, etc. and be done with it (maybe my editor gives me some hints). I feel like I would almost never do import a model (e.g., `allennlp_models.rc.models.naqanet`).

I recall in https://github.com/allenai/allennlp/issues/4051#issuecomment-610031589 , you mentioned that:

> I did it this way partially because the original plan was to have separate repositories for each of the categories, but also partially because when you work on a model, you usually work on the reader and the model at the same time, and it's confusing to have them far away from each other in the file hierarchy. It also makes it harder to copy and paste out exactly what you need.

It is unclear to me whether you expect people to be directly cloning the repo and using it, or importing things from it. In the former, I can see an argument for a _completely_ flat structure—it makes it easier for the user to browse around. I get the sense that the things that's "recommended" is the latter, though (using AllenNLP as a library). In that setting, I don't really care how close the files are to each other. I'm going to `pip install allennlp allennlp-models`, and import the components I need. If I need to copy/paste something, it's usually model code (just to skip all the boilerplate so i can modify `forward`), so I open up my web browser and navigate to `allennlp_models` and I again want to know exactly where it is.

> If these models are so core to the experience, why are they not in the core library?

You could argue that the model tools should live in the main repo, and only the models would be in this repo. To some extent, I feel like that would be ideal from an organizational perspective, but the tools are the biggest cause of the external dependencies, I think, so it'd be against the point of splitting things out in the first place...

matt-gardner:
A main issue is that there is a many-to-many mapping between models and datasets.  So there are re-usable components in here, and they need to be findable.  If you group by "model", how do I know which dataset is where?  If you group by "dataset", how do I know which model is where?  Given the many-to-many mapping, if you aim to have a library of useful things, no matter your perspective on why it exists, you need to impose some other structure than grouping by either models or datasets.

On what the difference is between this repo and the main one: the main repo has building blocks, this repo has things that are built with those building blocks.  Some of those things that are built are also themselves re-usable, hence there needs to be a discoverable structure, but that's the difference that I see.  It's a pretty clear conceptual divide, except in the case of metrics and other related tooling.  Those could end up in either place, but they seem to fit better in here to me.

dirkgr:
I believe this is everything then.

nelson-liu:
This looks great, thanks so much for all of your hard work! One thing that caught my eye while browsing the directories: some metrics call external scripts (e.g., the SQuAD EM / F1 metric, calls the SQuAD eval script, the SRL metric calls `srl-eval.pl`, etc). In the RC context, I noticed that you made a separate folder called `evaluations`, whereas in the SRL context `srl-eval.pl` is just in the `metrics` folder.

Maybe we should generalize this and put all such external files into a `tools` folder (e.g., like how EVALB is in `allennlp.tools` https://github.com/allenai/allennlp/tree/master/allennlp/tools ). On that thought, maybe `scripts` should go in here as well (e.g., the scripts in `structured_prediction`)?

PR #64, opened 2020-05-27
Use new Instance.duplicate() method instead of deepcopy

epwalsh:
Followup to https://github.com/allenai/allennlp/pull/4294.


requirements.txt
```diff
@@ -1,4 +1,6 @@
-git+https://github.com/allenai/allennlp.git@master
+# git+https://github.com/allenai/allennlp.git@master
+# TODO: remove
+git+https://github.com/epwalsh/allennlp.git@text-field-deep-copy
```
epwalsh, line 3: Need to revert this before merging.

PR #65, opened 2020-05-28
fixes on some structured prediction dataset readers

masashi-y:
Hello, thank you for sharing the great codebase! I am currently using the syntactic and semantic dependency parsers in this repo in my project and noticed some minor bugs.

- **incorrect `label_namespace` in `UniversalDependenciesDatasetReader`**: since `label_namespace` for `head_indices` (syntactic heads) is set to `head_index_tags`, these are treated as *tags*, resulting in the creation of the corresponding namespace in vocabulary with entries of numbers (head indices). Additionally, I observed inaccurate training results of a  biaffine model with this setting, while after fixing this, the training went good. I believe the `label_namespace`'s name for this used to be `head_indices`, so I suggest to return this back to the original name.

- **`skip_when_no_arcs` optional argument in `SemanticDependenciesDatasetReader`**: In the original implementation of this class, examples with no annotated semantic dependencies are discarded. This is fine for the training split, but it discards some examples in the validation/test sets, and results in inaccurate metric calculation. I suggest to add a new `skip_when_no_arcs` optional argument to control if to discard those examples, and set `validation_dataset_reader` with this key set false in the config json file.

Lastly I fixed the "initializer" part of `semantic_dependencies.json` to adapt to newer way of configuring this.

### DIFF
--- a/allennlp_models/structured_prediction/dataset_readers/universal_dependencies.py
+++ b/allennlp_models/structured_prediction/dataset_readers/universal_dependencies.py
@@ -111,7 +111,7 @@ class UniversalDependenciesDatasetReader(DatasetReader):
                 [x[0] for x in dependencies], text_field, label_namespace="head_tags"
             )
             fields["head_indices"] = SequenceLabelField(
-                [x[1] for x in dependencies], text_field, label_namespace="head_index_tags"
+                [x[1] for x in dependencies], text_field, label_namespace="head_indeices"
             )
 
         fields["metadata"] = MetadataField({"words": words, "pos": upos_tags})
--- a/allennlp_models/structured_prediction/dataset_readers/universal_dependencies.py
+++ b/allennlp_models/structured_prediction/dataset_readers/universal_dependencies.py
@@ -111,7 +111,7 @@ class UniversalDependenciesDatasetReader(DatasetReader):
                 [x[0] for x in dependencies], text_field, label_namespace="head_tags"
             )
             fields["head_indices"] = SequenceLabelField(
-                [x[1] for x in dependencies], text_field, label_namespace="head_indeices"
+                [x[1] for x in dependencies], text_field, label_namespace="head_index_tags"
             )
 
         fields["metadata"] = MetadataField({"words": words, "pos": upos_tags})
--- a/training_config/structured_prediction/semantic_dependencies.json
+++ b/training_config/structured_prediction/semantic_dependencies.json
@@ -2,7 +2,7 @@
     "dataset_reader":{
         "type":"semantic_dependencies"
     },
-    "dataset_reader":{
+    "validation_dataset_reader":{
         "type":"semantic_dependencies",
         "skip_when_no_arcs": false
     },
--- a/training_config/structured_prediction/semantic_dependencies.json
+++ b/training_config/structured_prediction/semantic_dependencies.json
@@ -2,6 +2,10 @@
     "dataset_reader":{
         "type":"semantic_dependencies"
     },
+    "dataset_reader":{
+        "type":"semantic_dependencies",
+        "skip_when_no_arcs": false
+    },
     "train_data_path": "/home/markn/data/semantic_dependency_parsing/semeval2015_data/dm/data/english/english_dm_augmented_train.sdp",
     "validation_data_path": "/home/markn/data/semantic_dependency_parsing/semeval2015_data/dm/data/english/english_dm_augmented_dev.sdp",
     "test_data_path": "/home/markn/data/semantic_dependency_parsing/semeval2015_data/dm/data/english/english_id_dm_augmented_test.sdp",
@@ -35,15 +39,18 @@
       "tag_representation_dim": 100,
       "dropout": 0.3,
       "input_dropout": 0.3,
-      "initializer": [
-        [".*feedforward.*weight", {"type": "xavier_uniform"}],
-        [".*feedforward.*bias", {"type": "zero"}],
-        [".*tag_bilinear.*weight", {"type": "xavier_uniform"}],
-        [".*tag_bilinear.*bias", {"type": "zero"}],
-        [".*weight_ih.*", {"type": "xavier_uniform"}],
-        [".*weight_hh.*", {"type": "orthogonal"}],
-        [".*bias_ih.*", {"type": "zero"}],
-        [".*bias_hh.*", {"type": "lstm_hidden_bias"}]]
+      "initializer": {
+        "regexes": [
+          [".*feedforward.*weight", {"type": "xavier_uniform"}],
+          [".*feedforward.*bias", {"type": "zero"}],
+          [".*tag_bilinear.*weight", {"type": "xavier_uniform"}],
+          [".*tag_bilinear.*bias", {"type": "zero"}],
+          [".*weight_ih.*", {"type": "xavier_uniform"}],
+          [".*weight_hh.*", {"type": "orthogonal"}],
+          [".*bias_ih.*", {"type": "zero"}],
+          [".*bias_hh.*", {"type": "lstm_hidden_bias"}]
+        ]
+      }
     },
 
     "data_loader": {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Nothing yet
+- Added option to SemanticDependenciesDatasetReader to not skip instances that have no arcs, for validation data.
 
 ## [v1.0.0rc5](https://github.com/allenai/allennlp-models/releases/tag/v1.0.0rc5) - 2020-05-14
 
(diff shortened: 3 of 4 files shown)


epwalsh:
Thanks @masashi-y! @matt-gardner you might be the best one to look at this

matt-gardner:
I can definitely take a look, just FYI @masashi-y that we have a paper submission deadline on Monday that's taking all of my time, so I'm not going to get to this until at least Tuesday.

dirkgr:
I pulled out the fix for the training config here: https://github.com/allenai/allennlp-models/pull/66

Regardless of what you decide with this PR, that's a fix we should have. Thanks @masashi-y!

masashi-y:
For a quick note, I noticed the second issue when I wanted to evaluate the trained graph_parser model on the test split of the SemEval2015, using the original [evaluation script](https://github.com/semantic-dependency-parsing/toolkit).
I have written the following `semantic_dependencies_predictor.py` to decode in the CoNLL-like format.
The output file contains less sentences than the input does, and the evaluation scripts outputs an error.

```py
from typing import Dict, Any, List, Tuple

from overrides import overrides
from io import StringIO

from allennlp.common.util import JsonDict, sanitize
from allennlp.data import DatasetReader, Instance
from allennlp.models import Model
from allennlp.predictors.predictor import Predictor
from allennlp.data.tokenizers.spacy_tokenizer import SpacyTokenizer


@Predictor.register("semantic_dependencies_predictor")
class SemanticDependenciesPredictor(Predictor):

    def __init__(
        self, model: Model, dataset_reader: DatasetReader, language: str = "en_core_web_sm"
    ) -> None:
        super().__init__(model, dataset_reader)
        self._tokenizer = SpacyTokenizer(language=language, pos_tags=True)

    def dump_line(self, outputs: JsonDict) -> str:
        OFFSET = 7
        words = outputs['words']
        arcs = outputs['arcs']
        arc_tags = outputs['arc_tags']
        predicates = {i: {} for i in range(len(words))}
        for (i, j), tag in zip(arcs, arc_tags):
            predicates[j][i] = tag
        predicates = {
            predicate: deps for predicate, deps in predicates.items()
            if len(deps) > 0
        }
        lines = [
            [
                str(i + 1),
                word,
                word.lower(),
                'POS',
                '-',
                '+' if i in predicates else '-',
                '_'
            ] + ['_' for _ in range(len(predicates))]
            for i, word in enumerate(words)
        ]
        for head, (_, deps) in enumerate(sorted(predicates.items())):
            for dep, tag in deps.items():
                lines[dep][OFFSET + head] = tag
        lines = '\n'.join('\t'.join(line) for line in lines)
        return f'#20000000\n{lines}\n\n'


    def predict(self, sentence: str) -> JsonDict:
        """
        Predict a semantic dependency parse for the given sentence.
        # Parameters

        sentence The sentence to parse.

        # Returns

        A dictionary representation of the dependency tree.
        """
        return self.predict_json({"sentence": sentence})

    @overrides
    def _json_to_instance(self, json_dict: JsonDict) -> Instance:
        """
        Expects JSON that looks like `{"sentence": "..."}`.
        """
        spacy_tokens = self._tokenizer.tokenize(json_dict["sentence"])
        sentence_text = [token.text for token in spacy_tokens]
        pos_tags = [token.tag_ for token in spacy_tokens]
        return self._dataset_reader.text_to_instance(sentence_text, pos_tags)

    @overrides
    def predict_instance(self, instance: Instance) -> JsonDict:
        outputs = self._model.forward_on_instance(instance)
        return sanitize(outputs)

    @overrides
    def predict_batch_instance(self, instances: List[Instance]) -> List[JsonDict]:
        outputs = self._model.forward_on_instances(instances)
        return sanitize(outputs)

```

training_config/structured_prediction/semantic_dependencies.json
```diff
@@ -2,6 +2,10 @@
     "dataset_reader":{
         "type":"semantic_dependencies"
     },
+    "dataset_reader":{
```
matt-gardner, line 5: Doesn't this need to be `validation_dataset_reader`?

allennlp_models/structured_prediction/dataset_readers/universal_dependencies.py
```diff
@@ -111,7 +111,7 @@ def text_to_instance(
                 [x[0] for x in dependencies], text_field, label_namespace="head_tags"
             )
             fields["head_indices"] = SequenceLabelField(
-                [x[1] for x in dependencies], text_field, label_namespace="head_index_tags"
+                [x[1] for x in dependencies], text_field, label_namespace="head_indeices"
```
matt-gardner, line 114: I dug into this trying to figure out what's going on here, and I'm quite confident that this label namespace is entirely ignored, because the input to the field is a list of integers.  The vocab namespace that's referenced here is never read in the model, and the field itself is never used as if it's a string.  It could alternatively have been a `ListField[IndexField]`, and that would be more accurate for what's actually going on here.

```suggestion
                [x[1] for x in dependencies], text_field, label_namespace="ignored"
```

allennlp_models/structured_prediction/dataset_readers/universal_dependencies.py
```diff
@@ -111,7 +111,7 @@ def text_to_instance(
                 [x[0] for x in dependencies], text_field, label_namespace="head_tags"
             )
             fields["head_indices"] = SequenceLabelField(
```
matt-gardner, line 113:
```suggestion
            # Semantically, using a ListField[IndexField] here would be more appropriate,
            # but we'll do it like because it's easier.
            fields["head_indices"] = SequenceLabelField(
```

matt-gardner:
Also, can you update to master and resolve the file conflict, and add a quick note to the changelog saying something like `Added option to SemanticDependenciesDatasetReader to not skip instances that have no arcs, for validation data.`

allennlp_models/structured_prediction/dataset_readers/universal_dependencies.py
masashi-y: After reading your reply, I checked this again and realized that this was my mistake. I was actually using a version of this code with a slight modification (which was the actual cause of the issue) and jumped to a wrong conclusion. I'm sorry for the trouble. I'd like it being `head_index_tags` rather than `ignored`, so I reverted the code.

training_config/structured_prediction/semantic_dependencies.json
masashi-y: Right, I fixed this!

masashi-y:
By the way, the semantic dependency parsing code has another issue that it's performance is unnecessarily poorer than other reported results for theSemEval 2015, since it does not consider edges for "top" relation (the 5th column in the [format](http://alt.qcri.org/semeval2015/task18/index.php?id=data-and-tools)), which is kind of semantic dependency equivalent of "root" relation of syntactic dependencies connecting the main verb to a dummy root token. The current `SemanticDependenciesDatasetReader` simply ignores this relation.

This can be fixed by adding space for the root token in the `AdjacencyField` objects, but the issue is that the class is a bit strict with regards to the input's length, that is, it requires that its inputs `arc_indices` and `token_field` be in accordance with each other in terms of the sentence's length.
The quickest fix is to every time extend the `token_field` with an additional dummy token at index 0, but this may not be ideal since it requires one extra space for the token in the vocabulary, and I think this goes against the allennlp's design choise as well (seeing that the biaffine parser's implementation also avoids this extra token).
The alternative fix is to make the `AdjacencyField` to have an option to be able to represent a adjacency matrix of the length of the sentence plus one. I think this is better since it helps to represent a syntactic dependency tree as well, and may be useful for some applications using that.

This is just to report, but if your are interested, I may be able to send another pull request, or make an issue at the main allennlp project page.

matt-gardner:
If you make that fix and it improves performance, we'd love to have a PR for it.

jbrry:
I used the `AdjacencyField` for parsing enhanced Universal Dependencies and I had to make the following changes to deal with an additional root token.

In my `dataset_reader`, I appended a 'root' token to `tokens` (which is just a list of the input tokens and not a TextField) so that it is sequence length + 1.
```
token_field_with_root = ['root'] + tokens
fields["enhanced_tags"] = RootedAdjacencyField(arc_indices, token_field_with_root, arc_tags, label_namespace="deps")
```

I made a custom `RootedAdjacencyField` which mimics the behaviour of the original but changes the behaviour of the [sequence_field](https://github.com/allenai/allennlp/blob/7d66b3e720cb3e1ffdd7a56e1b17ca23d91d03fd/allennlp/data/fields/adjacency_field.py#L61) value. I didn't use the `SequenceField` as I didn't want to add a dummy 'root' token to it (this was the behaviour in 2018 or so, e.g. [here](https://github.com/allenai/allennlp/commit/de0d3f7329c9c4f0204d87e76b6a956b0930ef3e) but it has since been removed). So instead, I made a token_sequence which is just a list of tokens.

```
+         token_sequence: List[str],
-         sequence_field: SequenceField,
```

In the original, the `sequence_field` can return its length which I was no longer able to do so I replaced this behaviour everywhere like in the below:
```
+         return {"num_tokens": len(self.token_sequence)}
-         return {"num_tokens": self.sequence_field.sequence_length()}
```

This seemed to do the trick but there may be better solutions!

PR #70, opened 2020-06-04
Fix capitalization typo in SNLI RoBERTa config

matt-gardner:
We aren't testing these at all, are we?  Is there a relatively easy way to make sure we're doing some minimal tests for these?

### DIFF
--- a/training_config/pair_classification/snli_roberta.jsonnet
+++ b/training_config/pair_classification/snli_roberta.jsonnet
@@ -8,7 +8,7 @@ local cls_is_last_token = false;
     "tokenizer": {
       "type": "pretrained_transformer",
       "model_name": transformer_model,
-      "add_special_tokens": False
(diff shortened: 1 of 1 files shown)


epwalsh:
> We aren't testing these at all, are we? Is there a relatively easy way to make sure we're doing some minimal tests for these?

I've thought about this a little. It would be really great if we had a quick, general way to validate configuration files.

matt-gardner:
> I've thought about this a little. It would be really great if we had a quick, general way to validate configuration files.

I had a simple suggestion here: https://github.com/allenai/allennlp/issues/3649#issuecomment-575926448.

epwalsh:
@matt-gardner should we make this a Milestone 1.1 priority?

matt-gardner:
I'll add that issue to the milestone, so we at least consider it.  Easy to move it out if we decide not to do it.

PR #75, opened 2020-06-17
Multiple Choice

dirkgr:
This has some tests now.


allennlp_models/mc/dataset_readers/fake.py
```diff
@@ -0,0 +1,89 @@
...
+logger = logging.getLogger(__name__)
+
+
+@DatasetReader.register("fake")
+class FakeReader(DatasetReader):
+    """
+    Creates fake multiple-choice input. If your model doesn't get 99% on this data, it is broken.
+
+    Instances have two fields:
+     * ``alternatives``, a ListField of TextField
```
epwalsh, line 16: Super super annoying nit-pick, but all of these double backticks should just be single backticks or they won't render properly in the docs when we address https://github.com/allenai/allennlp/issues/4124 😬

training_config/mc/commonsenseqa.jsonnet
```diff
@@ -0,0 +1,52 @@
...
+  },
+  "data_loader": {
+    "sampler": "random",
+    "batch_size": gpu_batch_size
+  },
+  "trainer": {
+    "optimizer": {
+      "type": "huggingface_adamw",
+      "weight_decay": 0.01,
+      "parameter_groups": [[["bias", "LayerNorm\\.weight", "layer_norm\\.weight"], {"weight_decay": 0}]],
```
epwalsh, line 30: I've seen this in some other configs and I'm just curious what the rational for it is?

dirkgr: There are two different spellings for the `LayerNorm.weight` parameters. HF changed from one to the other at some point. This captures both of them.

As for why we turn off weight decay for those parameters? I'm not sure. I think it comes from the original BERT paper where they recommend that. It makes some intuitive sense to me that you wouldn't want to put a decay on biases, but I don't know about layer norm.

dirkgr:
Fine-grained NER is broken again, but it's also broken in `master`, so I'll merge anyways.

PR #76, opened 2020-06-18
Make CopyNet work with pretrained transformer models

epwalsh:
Tests will fail until https://github.com/allenai/allennlp/pull/4377 is merged.

### DIFF
--- a/allennlp_models/generation/dataset_readers/copynet_seq2seq.py
+++ b/allennlp_models/generation/dataset_readers/copynet_seq2seq.py
@@ -1,16 +1,21 @@
 import logging
 from typing import List, Dict
+import warnings
 
 import numpy as np
 from overrides import overrides
 
-from allennlp.common.checks import ConfigurationError
 from allennlp.common.file_utils import cached_path
 from allennlp.common.util import START_SYMBOL, END_SYMBOL
 from allennlp.data.dataset_readers.dataset_reader import DatasetReader
 from allennlp.data.fields import TextField, ArrayField, MetadataField, NamespaceSwappingField
 from allennlp.data.instance import Instance
-from allennlp.data.tokenizers import Token, Tokenizer, SpacyTokenizer
+from allennlp.data.tokenizers import (
+    Token,
+    Tokenizer,
+    SpacyTokenizer,
+    PretrainedTransformerTokenizer,
+)
 from allennlp.data.token_indexers import TokenIndexer, SingleIdTokenIndexer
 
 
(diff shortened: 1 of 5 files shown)


allennlp_models/generation/dataset_readers/copynet_seq2seq.py
```diff
@@ -101,23 +115,41 @@ def __init__(
...
+        add_start_and_end_tokens: bool = True,
         **kwargs,
     ) -> None:
         super().__init__(**kwargs)
         self._target_namespace = target_namespace
         self._source_tokenizer = source_tokenizer or SpacyTokenizer()
         self._target_tokenizer = target_tokenizer or self._source_tokenizer
         self._source_token_indexers = source_token_indexers or {"tokens": SingleIdTokenIndexer()}
-        if "tokens" not in self._source_token_indexers or not isinstance(
-            self._source_token_indexers["tokens"], SingleIdTokenIndexer
```
epwalsh, line 112: Turns out this is really not necessary.

allennlp_models/generation/dataset_readers/copynet_seq2seq.py
```diff
@@ -72,6 +77,15 @@ class CopyNetDatasetReader(DatasetReader):
     source_token_indexers : `Dict[str, TokenIndexer]`, optional
         Indexers used to define input (source side) token representations. Defaults to
         `{"tokens": SingleIdTokenIndexer()}`.
+    add_start_and_end_tokens : `bool`, optional (default = `True`)
+        The `CopyNet` model assumes that the source sequence starts with a special
```
matt-gardner, line 81: Is this true?  I don't think it is.  You removed the index lookups from the model, and nothing broke.  These tokens were entirely unused, and if the user wants to have some additional special tokens like this on the start side for whatever modeling reason they want, they are easy to add in the tokenizer configuration.  I would just remove this parameter.

test_fixtures/generation/copynet/experiment_transformer.jsonnet
```diff
@@ -0,0 +1,69 @@
+local target_namespace = "target_tokens";
+local transformer_model = "bert-base-cased";
```
matt-gardner, line 2: There are smaller ones that you can use here, to make this faster.  E.g., see [here](https://github.com/allenai/allennlp-guide/blob/f2eb4c91f6576ecea639171d3a98f9e0846702a0/exercises/part2/representing-text-as-features/token_embedders_contextual.py#L29-L31).

allennlp_models/generation/dataset_readers/copynet_seq2seq.py
epwalsh: The model does make this assumption since adding special start/end tokens is a typical thing to do and the model needs to avoid computing "copy" scores for those special tokens so that there's no possibility of generating those tokens in a predicted target sequence.

epwalsh: "needs to" is probably too strong to say, but I think that was the rational.

matt-gardner: But the model doesn't actually do that.  `self._src_start_index` was never used in the model.

epwalsh: See here:
- https://github.com/allenai/allennlp-models/blob/master/allennlp_models/generation/models/copynet_seq2seq.py#L340
- https://github.com/allenai/allennlp-models/blob/master/allennlp_models/generation/models/copynet_seq2seq.py#L356

The encodings of the start and end tokens are sliced off in various places.

matt-gardner: Thanks, you're right, I missed that.  That's pretty brittle, though.  Might be nice to have that configurable somehow.  But probably better done in a separate PR.  For this one, I might consider at least leaving a note somewhere about how we always take off the first and last token.

epwalsh: Agreed. I'll follow up with a PR for that

test_fixtures/generation/copynet/experiment_transformer.jsonnet
epwalsh: I'm getting some weird errors when trying to use that model:

image

PR #77, opened 2020-06-21
Add SQuAD2 dataset reader + BiDAF no-answer model

nelson-liu:
Fixes https://github.com/allenai/allennlp/issues/1629

BiDAF-no-answer is a pretty antiquated model by now, so it would be nice to get something more modern in :) I don't have the bandwidth to work on that, though...I'm merely pushing up things that I had in private repos and never got around to contributing back upstream.

TODO:
- [x] can someone upload the SQuAD 2 train / dev sets to the AllenNLP AWS bucket?  Would be good to have them at https://allennlp.s3.amazonaws.com/datasets/squad/squad-train-v2.0.json and https://allennlp.s3.amazonaws.com/datasets/squad/squad-dev-v2.0.json , for consistency with squad 1.1
- [ ] I don't have the resources or time right now to train this model and turn it into a full-featured demo / pretrained model, can someone at AI2 help out with this? ( cc @dirkgr / @epwalsh ) . I added a config file that is just the BiDAF config file, with the minimal changes necessary for SQuAD 2.0 —it should work fine, but you can probably do better with some tuning.
- [ ] The only reason why the BidafNoAnswer model diff was so large was because i implemented a `no_answer_threshold` argument, though this isn't strictly necessary. The `bidaf` model would work fine with the squad2 dataset reader, since the only thing that changes is that a special no-answer token is appended to the input paragraph. Might be worth trying to train a better model on SQuAD 2.0 .
- [ ] At the moment, the SQuAD2 metric isn't totally accurate. The SQuAD2.0 metric, if you provide probabilities of each answer, will calculate the optimal no-answer threshold and score you based off of that. The right way to run it is to: (1) dump the predictions of the entire dataset, (2) run the squad2 metric on the entire dataset, picking the optimal no-answer threshold. I haven't implemented this here. HuggingFace transformers implements it like so : https://github.com/huggingface/transformers/blob/master/examples/question-answering/run_squad.py#L384-L401


dirkgr:
I uploaded the SQuAD 2.0 files. They were already there, but under a different name.

dirkgr:
I also kicked off a training process. Let's see where it lands.

dirkgr:
>  At the moment, the SQuAD2 metric isn't totally accurate. The SQuAD2.0 metric, if you provide probabilities of each answer, will calculate the optimal no-answer threshold and score you based off of that. The right way to run it is to: (1) dump the predictions of the entire dataset, (2) run the squad2 metric on the entire dataset, picking the optimal no-answer threshold. I haven't implemented this here.

I don't want to implement this here either. We don't really have a good system to run something after all the predictions are done. Instead, let's train a model, evaluate with that model, and set the threshold accordingly.

allennlp_models/rc/dataset_readers/squad2.py
```diff
@@ -0,0 +1,197 @@
...
+    and ``span_end``, both ``IndexFields`` into the ``passage`` ``TextField``.  We also add a
+    ``MetadataField`` that stores the instance's ID, the original passage text, gold answer strings,
+    and token offsets into the original passage, accessible as ``metadata['id']``,
+    ``metadata['original_passage']``, ``metadata['answer_texts']`` and
+    ``metadata['token_offsets']``. This is so that we can more easily use the official SQuAD 2.0
+    evaluation script to get metrics.
+
+    This differs from the SquadReader in that we append a special no-answer token to each
+    context. If the model's argmax span is this token, then we predict that the question is unanswerable,
+    given the provided context.
```
dirkgr, line 33: I haven't read the rest yet, but if this is the only difference, can't we make it a flag on the original reader?

allennlp_models/rc/dataset_readers/squad2.py
```diff
@@ -0,0 +1,197 @@
...
+    skip_invalid_examples: ``bool``, optional (default=False)
+        if this is true, we will skip those invalid examples
+    """
+
+    def __init__(
+        self,
+        tokenizer: Tokenizer = None,
+        token_indexers: Dict[str, TokenIndexer] = None,
+        passage_length_limit: int = None,
+        question_length_limit: int = None,
```
dirkgr, line 64: These types are `Optional[int]`.

dirkgr: In fact, all of those type are `Optional[...]`.

allennlp_models/rc/dataset_readers/squad2.py
```diff
@@ -0,0 +1,197 @@
...
+        if self.passage_length_limit is not None:
+            passage_tokens = passage_tokens[: self.passage_length_limit]  # type: ignore
+
+        passage_tokens_with_no_answer = passage_tokens + [  # type: ignore
+            Token(
+                text=SQUAD2_NO_ANSWER_TOKEN,
+                idx=passage_tokens[-1].idx + len(passage_tokens[-1].text) + 1,  # type: ignore
+                lemma_=SQUAD2_NO_ANSWER_TOKEN,
+            )
+        ]
```
dirkgr, line 141: Now `passage_tokens_with_no_answer` will be one token too long, won't it?

allennlp_models/rc/models/bidaf_no_answer.py
```diff
@@ -0,0 +1,386 @@
...
+        self,
+        vocab: Vocabulary,
+        text_field_embedder: TextFieldEmbedder,
+        num_highway_layers: int,
+        phrase_layer: Seq2SeqEncoder,
+        matrix_attention: MatrixAttention,
+        modeling_layer: Seq2SeqEncoder,
+        span_end_encoder: Seq2SeqEncoder,
+        dropout: float = 0.2,
+        no_answer_threshold: float = None,
```
dirkgr, line 75: If this is the same as regular BiDAF, except for this parameter, can we have just one class for both?

dirkgr: @nelson-liu I'm happy to make that change if you don't have time, but I'd like to know whether that will work before I do it.

dirkgr:
Just running the config gets to 59.6% F1. That's not very good yet.

allennlp_models/rc/models/bidaf_no_answer.py
nelson-liu: Yeah, it's the same, but `forward()` needs to be tweaked significantly...which is why I didn't subclass.

allennlp_models/rc/dataset_readers/squad2.py
nelson-liu: I chose to interpret the limit as the limit on valid tokens, but I'm fine with trimming off `limit + 1`.

nelson-liu:
From the SQuAD 2.0 paper https://arxiv.org/pdf/1806.03822.pdf , they got 59.8 EM and 62.6 F1 with BiDAF-No-Answer. Can you outline how you trained and evaluated the model? Since I don't think just straight `allennlp evaluate` would work.

allennlp_models/rc/dataset_readers/squad2.py
nelson-liu: Yeah, that would work too!

epwalsh:
Hey @nelson-liu, sorry this fell of the radar for a while. I just talked with the rest of the team and we decided that it's probably not worth adding the BiDAF no-answer model since it's not state-of-the-art and probably wouldn't see much use. Instead I'm going to focus on fine-tuning a transformer-based model to near state-of-the-art so that we can add it to the demo.

We definitely want a dataset reader for SQuAD 2.0 though. And ideally, we'd just have a single reader that works for both versions of SQuAD. I was thinking I would just take the reader you have implemented and adapt that to work with SQuAD 1.1, and for both transformer and non-transformer models.

Would you mind merging this PR into the `squad` branch instead of `master`? Then I can take it over from there.

nelson-liu:
Great! I changed the base of the branch, so now it targets squad.

epwalsh:
Thank you @nelson-liu!

PR #79, opened 2020-06-22
dont add START and END tokens to source in CopyNet

epwalsh:
This is a followup to our discussion here: https://github.com/allenai/allennlp-models/pull/76#discussion_r442545439

### DIFF
--- a/allennlp_models/generation/dataset_readers/copynet_seq2seq.py
+++ b/allennlp_models/generation/dataset_readers/copynet_seq2seq.py
@@ -1,10 +1,10 @@
 import logging
 from typing import List, Dict
-import warnings
 
 import numpy as np
 from overrides import overrides
 
+from allennlp.common.checks import ConfigurationError
 from allennlp.common.file_utils import cached_path
 from allennlp.common.util import START_SYMBOL, END_SYMBOL
 from allennlp.data.dataset_readers.dataset_reader import DatasetReader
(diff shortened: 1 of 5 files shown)


allennlp_models/generation/dataset_readers/copynet_seq2seq.py
```diff
@@ -130,29 +104,6 @@ def __init__(
...
-            and self._source_tokenizer._add_special_tokens
-            and self._add_start_and_end_tokens
-        ):
-            warnings.warn(
-                "Your source_tokenizer is a PretrainedTransformerTokenizer, but your dataset reader "
-                "has 'add_start_and_end_tokens' set to True. "
-                "This may result in errors or duplicate special tokens in your source sequences. ",
-                UserWarning,
-            )
-        if (
```
matt-gardner, line 144: This warning still seems reasonable.

allennlp_models/generation/dataset_readers/copynet_seq2seq.py
```diff
@@ -199,18 +150,18 @@ def text_to_instance(
         """
 
         tokenized_source = self._source_tokenizer.tokenize(source_string)
-        if self._add_start_and_end_tokens:
-            tokenized_source.insert(0, Token(START_SYMBOL))
-            tokenized_source.append(Token(END_SYMBOL))
+        if not tokenized_source:
+            # If the tokenized source is empty, it will cause issues downstream.
+            raise ConfigurationError(
```
matt-gardner, line 155: I don't think `ConfigurationError` is the right thing here.  We should probably move away from using it anywhere except in `FromParams` and `Registrable`, preferring `ValueError` or whatever other error is appropriate.  This one looks like a `ValueError` to me.

epwalsh: Good point!

PR #80, opened 2020-06-22
update skip conditions on constituency-parser-test

epwalsh:


epwalsh:
Going to merge this now since it's blocking

dirkgr:
Now we're only running this if the spacy version is old? I think it would be better if we updated the test so it runs when spacy is recent.

epwalsh:
I was planning on adding a different test for new spacy so that all recent versions are covered, I just didn’t have a chance yesterday.

PR #84, opened 2020-06-25
Fine grained NER

dirkgr:
After this, we'll update the demo.

Fixes https://github.com/allenai/allennlp/issues/4373.


dirkgr:
Tests are failing, but this is due to the test runner. Since the tests succeeded before and there is no new code in here, I'm merging.

PR #85, opened 2020-06-30
fixes for next_token_lm

epwalsh:


epwalsh:
Now I'm seeing this issue when trying to run the next_token_lm demo off of this branch:

image

allennlp_models/lm/predictors/next_token_lm.py
```diff
@@ -20,7 +20,10 @@ def predictions_to_labeled_instances(
     ):
         new_instance = instance.duplicate()
         token_field: TextField = instance["tokens"]  # type: ignore
-        mask_targets = [Token(target_top_k[0]) for target_top_k in outputs["words"]]
+        mask_targets = [
+            Token(target_top_k_text[0], text_id=target_top_id_id)
```
matt-gardner, line 24: `top_id_id` is a bit odd, but otherwise LGTM.

matt-gardner:
For the issue you're seeing, the weights must have changed somehow, either in huggingface code or in allennlp code.  I don't remember any model changes that should have caused this on our side, but I could be forgetting something.  To fix that, the easiest thing to do is to just instantiate the model in code and save it.  There are no parameters that need to be trained in that model, as it's just taking an existing pretrained model and putting it into a different format.

epwalsh:
Gotcha, thanks @matt-gardner

PR #89, opened 2020-07-08
implement manual distributed sharding for SNLI reader

epwalsh:
This gives a huge speed-up to SNLI/MNLI dataset reading in distributed training.


allennlp_models/pair_classification/dataset_readers/snli.py
```diff
@@ -45,6 +46,7 @@ def __init__(
         combine_input_fields: Optional[bool] = None,
         **kwargs,
     ) -> None:
+        kwargs["manual_distributed_sharding"] = True
         super().__init__(**kwargs)
```
dirkgr, line 50: I think it would be more readable to call `super().__init__(manual_distributed_sharding=True, **kwargs)`.

allennlp_models/pair_classification/dataset_readers/snli.py
```diff
@@ -60,20 +62,29 @@ def _read(self, file_path: str):
...
             logger.info("Reading SNLI instances from jsonl dataset at: %s", file_path)
-            for line in snli_file:
-                example = json.loads(line)
 
+        with open(file_path, "r") as snli_file:
+            example_iter = (json.loads(line) for line in snli_file)
+            filtered_example_iter = (
+                example for example in example_iter if example["gold_label"] != "-"
+            )
+            for example in itertools.islice(filtered_example_iter, start_index, None, step_size):
```
dirkgr, line 84: You could do the `islice` before the `json.loads`, so that it doesn't have to parse the JSON for lines that get discarded anyways. The problem is that then the filtering has to happen afterwards, which would mean different workers have different number of instances. This is not ideal, but not the end of the world.

I'm just leaving this here as a suggestion.

epwalsh: I had it that way initially but it didn't make a significant impact on speed.

PR #90, opened 2020-07-12
Update simple_seq2seq.py

wlhgtc:
1. support multi-layer decoders
2. return all top_k_predictions (tokens) for beam_search
3. support to load pre-train embedding files for target embedding


allennlp_models/generation/models/simple_seq2seq.py
```diff
@@ -66,21 +69,26 @@ class SimpleSeq2Seq(Model):
...
+            max_decoding_steps: int,
+            attention: Attention = None,
+            beam_size: int = None,
+            decoder_layers: int = 2,
+            target_pretrain_file: str = None,
+            target_namespace: str = "tokens",
+            target_embedding_dim: int = None,
+            scheduled_sampling_ratio: float = 0.0,
+            use_bleu: bool = True,
+            bleu_ngram_weights: Iterable[float] = (0.25, 0.25, 0.25, 0.25),
```
dirkgr, line 86: Can you add the new parameters at the end, so that the code stays backwards compatible?

allennlp_models/generation/models/simple_seq2seq.py
```diff
@@ -66,21 +69,26 @@ class SimpleSeq2Seq(Model):
...
     ) -> None:
         super().__init__(vocab)
+        self.source_embedding_dim = source_embedding_dim
```
dirkgr, line 89: Why no underscore before `_source_embedding_dim`?

dirkgr: Why did you add this parameter at all? Isn't it possible to get the source embedding dimension from the source embedder, without having to specify it? Also, if it's added here it needs to be added to the documentation.

allennlp_models/generation/models/simple_seq2seq.py
```diff
@@ -66,21 +69,26 @@ class SimpleSeq2Seq(Model):
...
-        bleu_ngram_weights: Iterable[float] = (0.25, 0.25, 0.25, 0.25),
+            self,
+            vocab: Vocabulary,
+            source_embedding_dim: int,
+            source_embedder: TextFieldEmbedder,
+            encoder: Seq2SeqEncoder,
+            max_decoding_steps: int,
+            attention: Attention = None,
+            beam_size: int = None,
+            decoder_layers: int = 2,
```
dirkgr, line 80: Should this be called `target_decoder_layers`, and default to 1?

allennlp_models/generation/models/simple_seq2seq.py
```diff
@@ -234,42 +239,46 @@ def forward(
...
-                indices = indices[: indices.index(self._end_index)]
-            predicted_tokens = [
-                self.vocab.get_token_from_index(x, namespace=self._target_namespace)
-                for x in indices
-            ]
-            all_predicted_tokens.append(predicted_tokens)
+            # if len(indices.shape) > 1:
+            #     indices = indices[0]
+            batch_predicted_tokens = []
+            for indices in top_k_predictions:
```
dirkgr, line 263: Why is the extra loop necessary now?

wlhgtc: The original code only return top1 result for beam_search. It's not convenient if we want eval top5 score (or we want choose result by some hand-craft algorithm). So I use the code segment in [copy-net](https://github.com/allenai/allennlp-models/blob/master/allennlp_models/generation/models/copynet_seq2seq.py#L866) to get all results.

allennlp_models/generation/models/simple_seq2seq.py
wlhgtc: Finish

allennlp_models/generation/models/simple_seq2seq.py
wlhgtc: Finish！

allennlp_models/generation/models/simple_seq2seq.py
wlhgtc: This parameter is useful when you have extra-features.
Suppose you have both word embedding(600 dim) and pos tags(600 dim). Then the embedding become (batch, length, 1200). But the encoder accept tensors with 600-dim. Then you may use a linear layer or directly add(just like style in BERT). This parameter will help.
But I have not finish all style "feature_merge" code in an elegant way.  **So I remove it.**
I will add some modules to finish it in the future.

wlhgtc:
@dirkgr All tests passed except a case with `target_decoder_layer = 2`. We have to update beam search code in [#4462](https://github.com/allenai/allennlp/pull/4462). Then it will pass.
But I wonder why I need **CI / Pretrained Models (pull_request)** test ? It has no relation with pretrained model.

dirkgr:
The pretrained tests were broken in `master` when you were doing this. The failure has nothing to do with you. You can fix it by merging from `master` again. Sorry about that.

wlhgtc:
@dirkgr  After thinking carefully about your advice. 
There are 2 different ways:
1.  Change code in beam_search, just as what I do in [allennlp/#4462](https://github.com/allenai/allennlp/pull/4462)
2. Adjust code in **all** LSMT decoder models in function `_forward_beam_search` :
https://github.com/allenai/allennlp-models/blob/4f2e316b47bb52c46e90c34efa1351223d219ee2/allennlp_models/generation/models/copynet_seq2seq.py#L546-L548
 https://github.com/allenai/allennlp-models/blob/4f2e316b47bb52c46e90c34efa1351223d219ee2/allennlp_models/generation/models/simple_seq2seq.py#L385-L387
 The code will looks like:
 ```
    # flatten operations
    ....

    self._beam_search.search(...)

    # unflatten operations
    ....

 ```
I prefer the first one, for the second one we need to repeat same code in many Models.

matt-gardner:
@wlhgtc, sorry, @dirkgr is out for a little bit, and we aren't sure if this is waiting on us or on you.  Can you give a quick update on where you think this stands?

wlhgtc:
@matt-gardner  Sorry, it's on my side. This PR aim to support multi-layer decoder in seq2seq model.
 I have finished the changes, but the code failed cause it needs some update in beam search(  [#4462](https://github.com/allenai/allennlp/pull/4462)).
Now this PR has been merged into master. 
But I don't know  the test code based on the allennlp code in master or  the newest released version ?
I will adjust my code and try it again in the next few days

dirkgr:
The code here will run against the latest master version of allennlp, but you might have to merge/resolve conflicts here before that happens.

wlhgtc:
@dirkgr  I re-push my code, all test case passed. 
But seem there is an  error about ssh key in [here](https://github.com/allenai/allennlp-models/pull/90/checks?check_run_id=956690323), could you help me fix it ?

allennlp_models/generation/models/simple_seq2seq.py
```diff
@@ -162,18 +178,19 @@ def take_step(
             the source mask, and the decoder hidden state and context. Each of these
             tensors has shape `(group_size, *)`, where `*` can be any other number
             of dimensions.
+<<<<<<< HEAD
+=======
         step : `int`
             The time step in beam search decoding.
 
+>>>>>>> 5d9098f6084a12da77b02d40e0d9392113aeb805
```
dirkgr, line 186: You checked in some unmerged files. I don't think it's serious, but we can't merge it like this.

tests/generation/predictors/seq2seq_test.py
```diff
@@ -19,7 +19,8 @@ def test_uses_named_inputs_with_simple_seq2seq(self):
         predicted_tokens = result.get("predicted_tokens")
         assert predicted_tokens is not None
         assert isinstance(predicted_tokens, list)
-        assert all(isinstance(x, str) for x in predicted_tokens)
+        for predicted_token in predicted_tokens:
+            assert all(isinstance(x, str) for x in predicted_token)
```
dirkgr, line 23: `predicted_tokens` is now a list of lists?

dirkgr:
I don't know about the ssh failure. @epwalsh, is it possible that this test can never succeed when the PR comes from a fork?

tests/generation/predictors/seq2seq_test.py
wlhgtc: Yes, it contains top n sequences, could see in [here](https://github.com/allenai/allennlp-models/pull/90#discussion_r453540252)

allennlp_models/generation/models/simple_seq2seq.py
wlhgtc: fixed

allennlp_models/generation/models/simple_seq2seq.py
```diff
@@ -45,7 +43,11 @@ class SimpleSeq2Seq(Model):
     target_embedding_dim : `int`, optional (default = `'source_embedding_dim'`)
         You can specify an embedding dimensionality for the target side. If not, we'll use the same
         value as the source embedder's.
-    attention : `Attention`, optional (default = `None`)
+    target_pretrain_file : `str` (default = None)
```
epwalsh, line 46:
```suggestion
    target_pretrain_file : `str`, optional (default = None)
```

epwalsh:
Just fixed the SSH issue with the docs. There was another build error in that job but it was because of bad formatting in a docstring. I think my suggestion would fix that though.

wlhgtc:
@epwalsh  Thanks for your advice. Now all tests pass, can we merge it into master ?

allennlp_models/generation/models/simple_seq2seq.py
dirkgr: Why did you resolve this? Isn't that parameter indeed optional?

I guess I would prefer it if the parameter was marked as `Optional[str]` too, but we don't do that consistently in the codebase yet, so I certainly don't expect you to fix it as part of this.

wlhgtc: Sorry, I don't understand what you mean. The `target_pretrain_file ` parameter is already optional and I just adjust the note part to pass the doc test.

dirkgr: @epwalsh suggested a change to mark the parameter as optional in the docs as well. It looks like you didn't take that suggestion? Or did I look at the wrong file?

dirkgr: Sorry, it looks like I looked at the wrong file. This change is already in.

dirkgr:
Thanks for sticking with it!

PR #91, opened 2020-07-13
build and publish models docs

epwalsh:
Closes https://github.com/allenai/allennlp/issues/4124. Published models docs to https://docs.allennlp.org/models/.


epwalsh:
> A lot of these things are copy and paste from the main repo?
> 
> A lot of the GitHub actions are duplicated. I assume we already talked about it on the main repo and decided this was the least of the available evils?

Yeup, a lot of copy-paste. We could potentially wrap up this logic into custom Actions, but it's probably not worth the effort at this point.

PR #92, opened 2020-07-14
Updated the fine-grained NER transformer model

dirkgr:
This is one of the models in https://github.com/allenai/allennlp/issues/4457.


dirkgr:
This model isn't in the demo, so no change required.

PR #98, opened 2020-07-22
Adding `ModelCard`

AkshitaB:
Fixes https://github.com/allenai/allennlp/issues/4203

This is currently work-in-prorgress.

Summary of changes:

The goal is to unify information of models in one place, and enforce standard ways to add new models.

* Adding a `ModelCard` class. Instances of this class are initiated for each model using the json files present in `modelcards` folder.
* `get_pretrained_models` method in pretrained.py reads all the json files present and constructs a dictionary of model objects of all pretrained models.

The following need to happen before PR is merged.
- [x] Add `ModelCard` and related classes.
- [x] Add tests.
- [x] Code cleanup.
- [x] Add `ModelCard` instances for each model in the repository.
- [ ] All models in demo should be present in pretrained.py

The following can be done later.
- [ ] Compare data with what is shown in demo.
- [ ] Fill model cards for all models.


allennlp_models/common/pretrained_model_config.py
```diff
@@ -0,0 +1,75 @@
...
+        self.description = description
+        self.last_trained_on = last_trained_on
+        self.predictor_name = predictor_name
+        self.overrides = overrides
+        self.model_card = model_card
+
+        self.__refs__[self.__class__].append(weakref.ref(self))
+
+    @staticmethod
+    def from_dict(config: Dict, model_class: Optional[Model] = None):
```
dirkgr, line 54: There are two ways of making one of these, one with all the parameters (`__init__()`), and one here. Maybe you can combine them by adding a `model_class` parameter to `__init__()`, with a default value, so you never need to pass a dict that's basically just a list of parameters anyways? Does the dict serve another purpose that I haven't glimpsed yet?

dirkgr: `model_class` should be `Optional[type]`. If it's `Optional[Model]`, that means it expects _instances_ of type `Model`.

I don't know if there is a way to restrict it more than `type`. It would be cool, but I don't know if that's possible in Python.

allennlp_models/common/pretrained_model_config.py
```diff
@@ -0,0 +1,75 @@
...
+def get_description(model_class):
+    """
+    Returns the model's description from the docstring.
+    """
+    return model_class.__doc__.split("# Parameters")[0].strip()
+
+
+@dataclass(frozen=True)
+class ModelCard(object):
+    pass
```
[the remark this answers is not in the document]
AkshitaB: I wanted to keep the ModelCard separate and cite it properly. And then the demo could just display its contents under the heading of "Model Card". But we can merge it with the rest of the config.

allennlp_models/common/pretrained_model_config.py
```diff
@@ -0,0 +1,75 @@
...
+        if self.archive_file and not self.archive_file.startswith("https:"):
+            self.archive_file = os.path.join(self._storage_location, self.archive_file)
+        self.name = name
+        self.description = description
+        self.last_trained_on = last_trained_on
+        self.predictor_name = predictor_name
+        self.overrides = overrides
+        self.model_card = model_card
+
+        self.__refs__[self.__class__].append(weakref.ref(self))
```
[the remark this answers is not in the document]
AkshitaB: Yes, I was iffy about weakref too, but was hesitating to use a global variable. I'll change it.

AkshitaB:
Summary of current status:

The following models in the demo don’t have a corresponding model/archive in allennlp_models:
* Reading comprehension
   - Nmn: drop_parser_bert
* Semantic parsing
   - Wikitables: wikitables_erm_parser
   - Nlvr: nlvr_coverage_parser
   - text to sql: atis_parser

The following models in allennlp_models do not have any archive files, and are not being demonstrated:
* Classification
   - BiattentiveClassificationNetwork
* Language Model
   - BidirectionalLanguageModel
* Generation
   - ComposedSeq2Seq
   - CopyNetSeq2Seq
   - SimpleSeq2Seq
* Pair_classification 
   - BiMpm
* Reading comprehension
   - BidafEnsemble
   - QaNet
* Structured prediction
   - GraphParser

dirkgr:
We don't have to have a demo for every model we have in the library. We don't have QaNet, but we have NaqaNet, so I'm not super concerned about it.

The first four models are in https://github.com/allenai/allennlp-semparse, but don't worry about bringing that repo up to date with your changes just yet.

allennlp_models/mc/models/transformer_mc.py
```diff
@@ -115,3 +117,23 @@ def get_metrics(self, reset: bool = False) -> Dict[str, float]:
...
+    id="mc-roberta-swag",
+    name="RoBERTa SWAG",
+    archive_file="swag.2020-07-08.tar.gz",
+    model_class=TransformerMC,
+)
+add_pretrained_model(
+    id="mc-roberta-commonsenseqa",
+    name="RoBERTa Common Sense QA",
+    archive_file="commonsenseqa.2020-07-08.tar.gz",
+    model_class=TransformerMC,
```
dirkgr, line 138: I hope I did a good job putting in a docstring for this ...

allennlp_models/common/model_card.py
```diff
@@ -0,0 +1,267 @@
+import os
+import logging
+from dataclasses import dataclass
+from typing import Optional, Union, Mapping, Dict
+
+from allennlp.models import Model
+from allennlp.common.checks import ConfigurationError
+
+PRETRAINED_MODELS: Dict = {}
```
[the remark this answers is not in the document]
dirkgr: By convention, having this name be all-uppercase means that we should consider it a constant, and it's fair game to access (but not change) from outside the module.

I think that's a defensible position to take, since this variable is definitely constant-ish, i.e., it should not change anymore after importing is done. I just want to point it out in case you didn't intend it that way.

allennlp_models/pretrained.py
```diff
@@ -1,194 +1,46 @@
...
+from allennlp_models.structured_prediction import models as sp_models  # noqa: F401
+from allennlp_models.tagging import models as tagging_models  # noqa: F401
 
+from allennlp_models.common.model_card import get_pretrained_models
 
-# flake8: noqa: E501
+# TODO: rename all archive files to follow standard naming.
+# Standard: task-model.yyyy-mm-dd.tar.gz OR model-task.yyyy-mm-dd.tar.gz
 
+PRETRAINED_MODELS = get_pretrained_models()
```
dirkgr, line 20: This is problematic. `model_card.py` might pick up more model cards after `pretrained.py` has been imported, but we've made a copy here, so we won't see those later updates anymore.

Why do we have to have this `dict` here anyways? Can you just read the needed information out of `model_card.py` directly in `get_predictor()`?

allennlp_models/pretrained.py
```diff
@@ -1,194 +1,46 @@
...
-    """
-    Reading Comprehension
-
-    Based on `BiDAF (Seo et al, 2017) <https://www.semanticscholar.org/paper/Bidirectional-Attention-Flow-for-Machine-Comprehen-Seo-Kembhavi/007ab5528b3bd310a80d553cccad4b78dc496b02>`_
-    """
-    predictor = _load_predictor(
-        "https://storage.googleapis.com/allennlp-public-models/bidaf-model-2020.03.19.tar.gz"
-    )
-    return predictor
+PRETRAINED_MODELS["roberta-sst"] = {"archive_file": "sst-roberta-large-2020.06.08.tar.gz"}
```
dirkgr, line 26: For these exceptions, is the problem that the models aren't available to treat like we treat the other models?

allennlp_models/common/model_card.py
```diff
@@ -0,0 +1,267 @@
...
+
+    def to_dict(self):
+        """
+        Only the non-empty attributes are returned, to minimize empty values.
+        """
+        info = {}
+        for key, val in self.__dict__.items():
+            if val:
+                info[key] = val
+        return info
```
dirkgr, line 54: Where do we consume these dicts? Can we just pass around the types you put in here?

allennlp_models/common/model_card.py
```diff
@@ -0,0 +1,267 @@
...
+    """
+
+    @classmethod
+    def from_object(cls, obj: Union[str, "ModelCardInfo"]):
+        """
+        Creates the relevant `ModelCardInfo` obj. If the input is str,
+        it is initialized as the first attribute of the class.
+        """
+        if isinstance(obj, str):
+            return cls(obj)  # type: ignore
```
dirkgr, line 43: Woha. So now that you have a base class that does this, you could add a `__str__` method that does the opposite, automatically. Something like
```Python
def __str__(self):
    if self.all_fields_are_str_or_none():
        return "\n\n".join(v for v in self.__dict__.values() if v is not None)
    else:
        return super(self).__str__()
```

allennlp_models/common/model_card.py
```diff
@@ -0,0 +1,267 @@
...
+        name = name or model_class.__name__
+        model_details = model_details or get_description(model_class)
+
+    if archive_file and not archive_file.startswith("https:"):
+        archive_file = os.path.join(STORAGE_LOCATION, archive_file)
+
+    config["name"] = name
+    config["archive_file"] = archive_file
+
+    config["model_details"] = ModelDetails.from_object(model_details)
```
matt-gardner, line 253: I haven't looked at this too closely, so feel free to ignore this if it doesn't make sense, but what's going on here reminds me a whole lot of what already happens in the `FromParams` logic. Is there an opportunity to reuse that here instead of writing this kind of code again?

allennlp_models/common/model_card.py
```diff
@@ -0,0 +1,267 @@
...
+                    info[key] = val
+        return info
+
+
+def add_pretrained_model(
+    id: str,
+    model_class: Optional[type] = None,
+    name: Optional[str] = None,
+    archive_file: Optional[str] = None,
+    overrides: Optional[Mapping] = None,
```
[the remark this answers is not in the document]
AkshitaB: Overrides are supposed to override the model's architecture defaults. I included them because of models like this one:

```{
    "id": "nmn",
    "archive_file": "https://storage.googleapis.com/allennlp-public-models/drop-nmn-2020.04.04.tar.gz",
    "predictor_name": "drop_demo_predictor",
    "overrides": {
        "model": {
            "debug": true,
            "beam_size": 1
       }
    }
}```

File: allennlp_models/pretrained.py (hunk 6 above)
Comment (AkshitaB) [reply allennlp_models/pretrained.py:20]: I created the dict envisioning that we could pass it to the demo directly.

File: allennlp_models/pretrained.py (hunk 7 above)
Comment (AkshitaB) [reply allennlp_models/pretrained.py:26]: Yes, that's right.

File: allennlp_models/common/model_card.py (hunk 8 above)
Comment (AkshitaB) [reply allennlp_models/common/model_card.py:54]: Related to the earlier question on having a dict in pretrained.py - the goal was to have an object that could easily be used in the demo.

File: allennlp_models/common/model_card.py (hunk 10 above)
Comment (AkshitaB) [reply allennlp_models/common/model_card.py:253]: I believe the `FromParams` logic is more for constructing the model object given the archive file (correct me if I'm wrong). The `ModelCard` is supposed to be more for information-purpose, where `model_details` are to store meta-information about the model (when it was trained, who developed it, etc). I have included the `archive_file` because we want to pass all this information to the demo.

File: allennlp_models/common/model_card.py [hunk 12]
```diff
@@ -0,0 +1,267 @@
...
+
+def get_description(model_class):
+    """
+    Returns the model's description from the docstring.
+    """
+    return model_class.__doc__.split("# Parameters")[0].strip()
+
+
+@dataclass
+class ModelCardInfo(object):
```
matt-gardner, line 29: Python 3 nit: inheriting from object is unnecessary in python 3.
```suggestion
class ModelCardInfo:
```

allennlp_models/pair_classification/models/bimpm.py
```diff
@@ -246,3 +249,6 @@ def get_metrics(self, reset: bool = False) -> Dict[str, float]:
         }
 
     default_predictor = "textual_entailment"
+
+
+add_pretrained_model(id="pc-bimpm", name="Bilateral Multi-Perspective Matching Model")
```
matt-gardner: It seems odd to me to have these live in the model implementation classes, instead of in a central place that has all of the pretrained models.  Conceptually that means I'm modifying model code if I just want to add a new pretrained model, instead of modifying a central list of pretrained models.

matt-gardner: I also don't really understand what we gain by having a "pretrained model" added without any weights.  What is this doing?

allennlp_models/common/model_card.py
matt-gardner: `FromParams` is a general framework for constructing objects from JSON dictionaries.  The reason I thought this was similar was because it automatically handles the case where your type is `Union[str, SomeObject]`.  What you are doing here is constructing a `ModelCard` object while handling union types, and then adding it to the list of pretrained models.  You could imagine instead having the constructor be separate, and having `FromParams` worry about handling these union types, instead of duplicating that logic in all of these classes.

This would let you, e.g., add a pretrained model to our list by just writing a JSON file with model card info, including a path to the archive file, and then we load whatever is in some directory using `FromParams` and can display them wherever we want.  This also has the benefit of being able to handle uploading these JSON files to a server somewhere, without needing any code modifications.  We could someday support a fully-automated third-party model upload site.

allennlp_models/common/model_card.py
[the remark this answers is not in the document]
dirkgr: We apply those values to the config after it is loaded from the archive? I didn't know we had that capability, but I immediately know what to use it for :-)

allennlp_models/pretrained.py
dirkgr: Couldn't the demo just call `get_pretained_models()`? Or just access `model_card.PRETRAINED_MODELS`?

allennlp_models/pretrained.py
dirkgr: Hmm, I guess we could just call `add_pretrained_model()` in here then, to keep the lists in sync.

But I'd much rather have only one list. Keeping two if them in sync is never fun.

allennlp_models/pair_classification/models/bimpm.py
AkshitaB: This is removed in the latest code commit. I was checking what models we have the pretrained archives for.

allennlp_models/common/model_card.py
dirkgr: I think this case is different. `FromParams` treats a single string `s` the same as `{"type": s}`, but this takes the single string and treats it as the first parameter to the constructor.

allennlp_models/pair_classification/models/bimpm.py
dirkgr: Conceptually, when you change the model code, you should also re-train and update the trained weights.

It makes sense to me either way.

allennlp_models/common/model_card.py
AkshitaB: Regarding the decision to include the config in python files instead of JSON files, @dirkgr and I spoke about this earlier and decided that anyone updating the json file can also update the python file. The current advantage to keeping it in code is that we pull some of the missing information from the `Model` object without having to manually fill it in. I am open to making changes if we decide otherwise.

AkshitaB: @matt-gardner I went through the `FromParams` logic, and I agree with @dirkgr . Since we want to provide an easy way to include some, if not all of the model card information, the goal of the base class `ModelCardInfo` and its method `from_object` to map the provided str argument to the first argument of the constructor. If I want to use FromParams, I would still need to reimplement the `from_params` method to do this, and I do not think that class is providing any additional functionality for this use case. Note: I am looking at the way `FromParams` is used in `Model`, but perhaps there is a more generalized way to use it that I am not grasping.

allennlp_models/pretrained.py
AkshitaB: Yes, it can. I wrote it this way so that pretrained.py remains the source of truth for everything outside, which is how it was originally meant to be. I put the global `PRETRAINED_MODELS` in `ModelCard` rather than pretrained.py only to simplify adding new pretrained models in the `ModelCard`'s init function itself. But we can technically do away with pretrained.py completely.

allennlp_models/common/model_card.py
matt-gardner: Just to be clear on what I'm suggesting:

```python
class ModelCard(FromParams):
    def __init__(
        self,
        model_class: Optional[type] = None,
        ...
        model_details: Optional[Union[str, ModelDetails]] = None,
        ...
    ):
        ...
        if isinstance(model_details, str):
            # Transparent logic for what happens when you're given a string, that can be documented
            # in this class's docstring.
            model_details = ModelDetails(description=str)
        self.model_details = model_details
        ...

def add_pretrained_model(model_card: ModelCard):
    # This method now looks kind of unnecessary
    PRETRAINED_MODELS[model_card.name] = model_card
    return model_card
```

And then instead of having a bunch of calls to `add_pretrained_model` scattered around the repo, we have a single directory that has json files describing model cards for each of our pretrained models, with a single method that finds all of those json files, and calls, essentially `add_pretrained_model(ModelCard.from_params(json_file))`.  Nothing else in your code needs to change, except that `ModelCardInfo` can go away, because it's no longer necessary.

The benefit over what you have here is that the constructor logic lives in one place, instead of two places.  If you want to add a parameter to `ModelCard`, with the code as it is now you have to modify both the `ModelCard` class and the `add_pretrained_model` method.  This is basically _exactly_ the situation we were in with our initial version of `from_params`, before we read type annotations to automatically discover and construct parameters.  It also seems conceptually simpler; it's not clear why the constructor for `ModelCard` should actually live inside of `add_pretrained_model`.  

One minor hiccup is that the data classes will also need to inherit from `FromParams` currently, though it seems easy enough to just make `FromParams` work automatically on data classes, as they will always have type annotations.

AkshitaB:
As per discussion with @matt-gardner and @dirkgr, the model card information has been moved to json files in the `modelcards` directory. pretrained.py simply has a function to get a dict of pretrained models, which can be called in the demo.

allennlp_models/modelcards/tagging-fine-grained-crf-tagger.json
```diff
@@ -0,0 +1,7 @@
+
+
+{
+    "name": "tagging-fine-grained-crf-tagger",
```
matt-gardner, line 4: Looking at this, I don't see why the "name" has to include all of this information.  If we rename this to `registered_model_name`, also, then it's more clear that you don't need to do this.

Oh, I see, below you are wanting to use this as the primary key for storing these objects.  Seems better to just not conflate the registered model name and the model card id.  Maybe add a new field for `registered_model_name`, or just `model_name`, keep the `name` as it is (or rename it to `id`, or something), and then you don't have to add all of the new calls to `Model.register`?

allennlp_models/pretrained.py
```diff
@@ -1,194 +1,32 @@
...
+    from allennlp_models.generation import models as generation_models  # noqa: F401
+    from allennlp_models.lm import models as lm_models  # noqa: F401
+    from allennlp_models.mc import models as mc_models  # noqa: F401
+    from allennlp_models.pair_classification import models as pc_models  # noqa: F401
+    from allennlp_models.rc import models as rc_models  # noqa: F401
+    from allennlp_models.structured_prediction import models as sp_models  # noqa: F401
+    from allennlp_models.tagging import models as tagging_models  # noqa: F401
 
+    pretrained_models = {}
+    modelcards = os.path.join(os.path.dirname(os.path.realpath(__file__)), "modelcards/*.json")
```
matt-gardner, line 27: Nit: maybe give this a more clear name, like `model_card_filenames`?

allennlp_models/pretrained.py
```diff
@@ -1,194 +1,32 @@
...
-    )
 
+    # These imports are included so that the model cards can be filled with default information
+    # obtained from the registered model classes.
 
-def roberta_commonsenseqa() -> TransformerMCPredictor:
-    return _load_predictor(
-        "https://storage.googleapis.com/allennlp-public-models/commonsenseqa.2020-07-08.tar.gz"
-    )
+    from allennlp_models.classification import models as classification_models  # noqa: F401
```
matt-gardner, line 16: You don't need the `as` statements here, do you?

allennlp_models/common/model_card.py
```diff
@@ -0,0 +1,276 @@
...
+    treated as the first argument of the relevant constructor.
+
+    """
+
+    _storage_location = "https://storage.googleapis.com/allennlp-public-models/"
+
+    def __init__(
+        self,
+        name: str,
+        model_class: Optional[type] = None,
```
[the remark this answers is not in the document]
AkshitaB: Yes, I was considering adding a function that loops over the list of available models, so that the model_class can be specified using a string name like `SimpleTagger`. I think Registrable only offers registered_name --> modulepath.ModelType right now.

allennlp_models/pretrained.py
```diff
@@ -1,194 +1,34 @@
...
 
+    # These imports are included so that the model cards can be filled with default information
+    # obtained from the registered model classes.
 
-def roberta_piqa() -> TransformerMCPredictor:
-    return _load_predictor(
-        "https://storage.googleapis.com/allennlp-public-models/piqa.2020-07-08.tar.gz"
+    pretrained_models = {}
+    model_card_paths = os.path.join(
+        os.path.dirname(os.path.realpath(__file__)), "modelcards/*.json"
```
dirkgr, line 28: If you want to do it this way, you have to make sure those json files are included in the wheel that `setup.py` builds.

allennlp_models/common/model_card.py
```diff
@@ -0,0 +1,282 @@
...
+        for key, val in self.__dict__.items():
+            if val:
+                info[key] = val
+        return info
+
+    def __str__(self):
+        display = ""
+        for key, val in self.to_dict().items():
+            display += "\n" + key.replace("_", " ").capitalize() + ": "
+            display += "\n\t" + val.replace("\n", "\n\t") + "\n"
```
dirkgr, line 35: This will mean that the final string starts and ends with `"\n"`. Return `display.strip()`?

allennlp_models/common/model_card.py
```diff
@@ -0,0 +1,282 @@
...
+        self.metrics = metrics
+        self.evaluation_data = evaluation_data
+        self.training_data = training_data
+        self.quantitative_analyses = quantitative_analyses
+        self.ethical_considerations = ethical_considerations
+        self.caveats_and_recommendations = caveats_and_recommendations
+
+    def to_dict(self) -> Dict[str, Any]:
+        """
+        Coverts the `ModelCard` to a flat dictionary object. This can be converted to
```
dirkgr, line 272: "Converts"

allennlp_models/pretrained.py
```diff
@@ -1,194 +1,34 @@
...
-
-
-def roberta_snli() -> DecomposableAttentionPredictor:
-    return _load_predictor(
-        "https://storage.googleapis.com/allennlp-public-models/snli_roberta-2020.06.09.tar.gz",
-        "textual-entailment",
-    )
 
+    # These imports are included so that the model cards can be filled with default information
+    # obtained from the registered model classes.
```
dirkgr, line 24: I think you removed the imports?

setup.py
```diff
@@ -87,6 +87,9 @@ def fix_url_dependencies(req: str) -> str:
     packages=find_packages(exclude=["*.tests", "*.tests.*", "tests.*", "tests"],),
     install_requires=install_requirements,
     include_package_data=True,
+    package_data={
+        "allennlp_models": ["modelcards/*.json", "structured_prediction/tools/srl-eval.pl"]
+    },
```
dirkgr, line 92: :+1:

PR #109, opened 2020-08-05
Update graph parser metrics

AkshitaB:
Should be merged along with https://github.com/allenai/allennlp/pull/4525 .


AkshitaB:
> LGTM, but doesn't pass tests? It will only pass with the other change at the same time, I assume?

Yes, that's right.

PR #112, opened 2020-08-12
validate pretrained configs in CI

epwalsh:
closes https://github.com/allenai/allennlp/issues/3649 by making sure every part of each training config can be instantiated. This CI job (called "Training Configs") will only run on master pushes and nightly builds to keep our CI snappy on pull requests.

PR #115, opened 2020-08-17
run config tests in subprocesses

epwalsh:
The training config tests I added in https://github.com/allenai/allennlp-models/pull/112 take up a ton of memory when they are all ran together, which can cause our self-hosted GitHub Actions runner to run out of memory.

There are probably several reasons why this is happening, some more obvious than others. One reason I can think of is that each different transformer that is loaded is kept in memory until pytest finishes due to the transformer caching mechanism we have.

Rather than try to hack around this particular cause, and others, I thought the easiest solution would be to run every one of these config tests in a separate forked subprocess. Luckily there is already a little pytest plugin that does just that.


epwalsh:
@dirkgr looks like we could by using `pytest-xdist` (https://pypi.org/project/pytest-xdist/). Might be worth it.

epwalsh:
@dirkgr okay I added xdist to run these in parallel and also made the same changes for the pretrained model tests.

epwalsh:
> Cool! How much faster is it?

For the "Training Configs" tests, it went from taking about 5 mins 20 seconds on my machine to 3 mins 30 seconds 👍

PR #118, opened 2020-08-19
always run configs CI job

epwalsh:
Now that we're running the config tests in parallel, they don't take nearly as long, so I think it's worth running them on PRs again.

This also fixes a bug I introduced on one of the last PRs where the pretrained model tests were being run in the "Checks" workflow, which makes that workflow take like 16 minutes 😮


dirkgr (approved):
👽

PR #122, opened 2020-08-25
improve generation dataset readers

epwalsh:
Related to https://github.com/allenai/allennlp/pull/4597, this makes the `generation` dataset readers efficient for distributed and multi-process data loading.

This should probably done for all of our readers, I just figured I would do them bit by bit.


allennlp_models/generation/dataset_readers/cnn_dm.py
```diff
@@ -157,15 +159,33 @@ def text_to_instance(
         if self._source_max_tokens is not None and len(tokenized_source) > self._source_max_tokens:
             tokenized_source = tokenized_source[: self._source_max_tokens]
 
-        source_field = TextField(tokenized_source, self._source_token_indexers)
+        source_field = TextField(
+            tokenized_source,
+            # Token indexers are applied later during multi-process loading with
+            # the `apply_token_indexers` method, so we only apply them now if there
+            # is a single worker.
+            None if self._worker_info is not None else self._source_token_indexers,
```
dirkgr, line 167: Can't we always do it later then? Does `apply_token_indexers` not run in the non-distributed setting?

epwalsh: Good question. `apply_token_indexers` is always called from the `DataLoader`, regardless of how many workers are being used. So it makes no difference during training.

But if we didn't have this logic here - i.e. always left the indexers out of the `Instance`s created by `text_to_instance()` - this would cause issues when someone implements a `Predictor` that utilizes this `DatasetReader` if they're not aware of this functionality and forget to call `apply_token_indexers`.

dirkgr: That seems preferable to having to do this big if statement in all the readers? I mean, that's also something you can get wrong when you write your own stuff, right?

Is there no place where we can call `apply_token_indexers` for them in the case of the predictor?

epwalsh: Hmm I'll take another look.

epwalsh: Ok, have another look at https://github.com/allenai/allennlp/pull/4597. Once that gets merged I think we remove the "if" conditions here.

allennlp_models/generation/dataset_readers/cnn_dm.py
```diff
@@ -157,15 +159,21 @@ def text_to_instance(
         if self._source_max_tokens is not None and len(tokenized_source) > self._source_max_tokens:
             tokenized_source = tokenized_source[: self._source_max_tokens]
 
-        source_field = TextField(tokenized_source, self._source_token_indexers)
+        source_field = TextField(tokenized_source)
```
dirkgr, line 162: I'm digging this, the fact that you don't even specify an indexer here anymore. That's where we want to go with the "Instances are nothing but data." concept.

dirkgr (approved):

PR #124, opened 2020-08-26
Fixes the BERT SRL model

dirkgr:
This is training right now on allennlp-server3. It should fix https://github.com/allenai/allennlp/issues/4392.


allennlp_models/structured_prediction/dataset_readers/srl.py
```diff
@@ -133,7 +133,13 @@ def __init__(
         **kwargs,
     ) -> None:
         super().__init__(**kwargs)
-        self._token_indexers = token_indexers or {"tokens": SingleIdTokenIndexer()}
+        if token_indexers is not None:
+            self._token_indexers = token_indexers
+        elif bert_model_name is not None:
+            from allennlp.data.token_indexers import PretrainedTransformerIndexer
+            self._token_indexers = {"tokens": PretrainedTransformerIndexer(bert_model_name, "tokens")}
```
matt-gardner, line 140: Are you sure you want to use "tokens" as the vocab namespace here?

matt-gardner: I'm assuming it's because of an assumption made in the `srl_bert.py` code.  But, e.g., I just wrote some stuff in the multi-task code that follows the assumption made in `PretrainedTransformerIndexer` that the default namespace is "tags".  It'd be better if we just modified all first-party code to make the same assumption.  I'd be in favor of changing `srl_bert.py` instead of changing the default value here.

dirkgr: No, but `SingleId` uses it.

dirkgr: I don't understand why the default is `"tags"`. Do you?

tests/pretrained_test.py
```diff
@@ -112,30 +110,28 @@ def test_semantic_role_labeling(self):
...
+                    "O",
+                    "O",
                     "B-V",
-                    "B-ARG1",
-                    "I-ARG1",
-                    "I-ARG1",
-                    "I-ARG1",
-                    "I-ARG1",
-                    "I-ARG1",
-                    "I-ARG1",
```
dirkgr, line 138: This is arguably a regression. We can't guarantee that our models won't ever regress on test cases when we re-train them, so I'll merge this anyways.

matt-gardner: I would classify this as an improvement, actually.  "Will" here is a modal verb, it shouldn't have any arguments on its own.  It's a failure of the heuristic in the predictor that this gets an entry in the result list at all.

PR #128, opened 2020-09-01
BART based (passage, answer) --> question generation model 

nitishgupta:
Given a passage with a span marked as the answer, fine-tune a BART model to generate a relevant question.

The dataset-reader is specific to SQuAD but the model is dataset agnostic.

PR contains a dataset reader, model, and a predictor.


nitishgupta:
@matt-gardner The test for the model fails as it cannot find the dataset-reader registered with "squad_question_generation" which is present and registered. Regular training with that reader works. Would you know what the issue might be?

matt-gardner:
On the registration error: I think that modifying the `__init__.py` files in the relevant modules should be enough.  You need to import your classes there, so they actually get registered.  I'm not sure why "regular training" succeeds for you locally, though.

matt-gardner (commented):
I haven't looked too closely at the model yet; what exactly had to change from the other BART model that we have?

nelson-liu:
this looks useful, thanks @nitishgupta ! Any plans on addressing Matt's comments in the near future (np if not, just wondering)?

matt-gardner:
ICLR deadline is Friday morning; I'd guess work on this will pick up a bit after that.

nitishgupta:
Yes, I plan to get back to this then.

PR #129, opened 2020-09-03
allow multiple token embedders, but only use first with non-empty type

tpanza:
Attempt to fix allenai/allennlp#4554


allennlp_models/lm/modules/token_embedders/language_model.py
```diff
@@ -76,15 +77,28 @@ def __init__(
...
+            # We need at least one non-empty embedder in the language model
+            raise ConfigurationError(
+                f"Language model from {archive_file} trained with only empty embedders!"
+            )
+        elif len(non_empty_embedders) > 1:
+            warnings.warn(
+                "Multiple non-empty embedders were found, but we will only use "
+                f"the first one: {non_empty_embedders[0]}",
+                RuntimeWarning,
+            )
```
dirkgr, line 99: This scenario will not work right. If the LM was trained with multiple embedders, it will need to be run with multiple embedders. It won't work to just leave one out.

allennlp_models/lm/modules/token_embedders/language_model.py
```diff
@@ -76,15 +77,28 @@ def __init__(
...
-            # Note: We only care about embedded indices. This does not include "tokens" which
-            # is just used to compute the loss in LanguageModel.
-            raise ConfigurationError(f"LM from {archive_file} trained with multiple embedders!")
-        self._token_name = token_names[0]
+        non_empty_embedders = [
+            token_embedder_key
+            for token_embedder_key, token_embedder_value in text_field_embedder[
+                "token_embedders"
+            ].items()
+            if token_embedder_value.get("type") != "empty"
```
dirkgr, line 85: This is somewhat problematic because we're basically duplicating the logic that parses the config files. There are other ways to specify an empty token embedder than this one.

Maybe we can just parse it using that same logic. I'm pretty sure you can just say `text_field_embedder = TextFieldEmbedder.from_params(Params(text_field_embedder))`, and it will create a `TextFieldEmbedder` object according to whatever the current config parsing rules are. Then you can do something like
```
non_empty_embedders = [
    name
    for name, token_embedder in text_field_embedder._token_embedders.items()
    if not isinstance(token_embedder, EmptyEmbedder)
]
```

If you want to be extra correct, you should check that `text_field_embedder` is of type `BasicTextFieldEmbedder`, though at the moment we only have that one anyways.

It's a little unfortunate that we have to read out the internal member of `BasicTextFieldEmbedder`, but it's less hacky than second-guessing the config file parser.

dirkgr (changes requested):

dirkgr:
@matt-gardner, do you have the background to look at these changes? I'm learning about the BidirectionalLanguageModel stuff as I go along here.

matt-gardner (approved):
I know why the empty embedder was used, but I don't have a whole lot of context on using the LM as a TokenEmbedder; I'm not sure that was ever really used for anything that we did.

Long term, I would probably vote for deprecating this code.  I think the `Backbone` stuff that I have in https://github.com/allenai/allennlp/pull/4601 would accomplish the same goal, but more easily.  As a short-term patch to the existing code, the solution that's in this PR seems correct to me.

dirkgr:
I would really like to validate this with an end-to-end run of training Transformer-ELMo, and then using it as an embedder, but it's taking a long time, so I'll merge this right now and follow up with other fixes if necessary.

PR #131, opened 2020-09-11
Transformer ELMo config fixes

dirkgr:
I don't know why these broke, since we have tests for the configs not breaking.


dirkgr:
This PR is now showing a random selection of commits, it seems, thanks to squash merges.

epwalsh:
Based on our discussion in standup, should I review this now or hold off?

dirkgr:
I don't remember what we talked about during standup 6 weeks ago, but I think this is ready for review.

Makefile
```diff
@@ -50,7 +50,7 @@ lint :
 
 .PHONY : format
 format :
-	black --check .
+	black --exclude training_config_test.py --check .
```
epwalsh, line 53: We could use

```python

# fmt: off
...
# fmt: on
```

comments within the code instead. I think that would be a better solution since many people might have black run automatically within their code editor (like me). Otherwise I would have to remember to turn it off everytime I edit this file.

dirkgr: I'll make the change.

epwalsh (approved):

PR #134, opened 2020-09-23
SQuAD 2.0 support

epwalsh:
This is ready for a review, though I still have a few things on my TODO list:

- [x] train baseline `TransformerQA` on official SQuAD v2.0 dataset with good results. I'll be putting experiments in this Beaker workspace: https://beaker.org/ws/ai2/SQuADv2.0/experiments?archived=false&sort=created:descending
    - [x] try RoBERTa model
    - [x] try with AMP
- [x] improve efficiency of forward pass of `TransformerQA` (no need to calculate EM and F1 on a training forward pass)
- [ ] allow different (more advanced) heads on `TransformerQA`, such as one that predicts the span end index based on additional context from the span start index (instead of predicted span start and span end independently). I'll probably do this in a separate PR.

----

## Updates

I got pretty good performance with `bert-base-cased` on this run: https://beaker.org/ex/ex_mjtl4hbndnkr/tasks/tk_c3bxbdsev5bh. I've uploaded this to Google Cloud: `gs://allennlp-public-models/transformer-qa-2020-10-01.tar.gz`. You can test it out with this:

```python
from allennlp.predictors import Predictor
from allennlp_models.rc import *

predictor = Predictor.from_path("https://storage.googleapis.com/allennlp-public-models/transformer-qa-2020-10-01.tar.gz")

passage = (
    "The Normans (Norman: Nourmands; French: Normands; Latin: Normanni) were "
    "the people who in the 10th and 11th centuries gave their name to Normandy, a region in France. "
    'They were descended from Norse ("Norman" comes from "Norseman") raiders and pirates from Denmark, '
    "Iceland and Norway who, under their leader Rollo, agreed to swear fealty to King Charles III of West Francia. "
    "Through generations of assimilation and mixing with the native Frankish and Roman-Gaulish populations, "
    "their descendants would gradually merge with the Carolingian-based cultures of West Francia. "
    "The distinct cultural and ethnic identity of the Normans emerged initially in the first half "
    "of the 10th century, and it continued to evolve over the succeeding centuries."
)

question = "In what country is Normandy located?"
predictor.predict(question, passage)
#> 'France'

question = "Who gave their name to Normandy in the 1000's and 1100's"
predictor.predict(question, passage)
#> ''  no answer
```


allennlp_models/modelcards/rc-transformer-qa.json
```diff
@@ -3,16 +3,16 @@
     "registered_model_name": "transformer_qa",
     "registered_predictor_name": null,
     "display_name": "Transformer QA",
-    "archive_file": "transformer-qa-2020-05-26.tar.gz",
+    "archive_file": "transformer-qa-2020-08-01.tar.gz",
```
dirkgr, line 6: `08-01` is August first, two months ago. Is that really the date on that model?

allennlp_models/rc/dataset_readers/transformer_squad.py
```diff
@@ -20,23 +20,39 @@
...
-       ``metadata['question']``, ``metadata['context']``, ``metadata['question_tokens']``,
-       ``metadata['context_tokens']``, and ``metadata['answers']. This is so that we can more easily use the
+     * `metadata`, a `MetadataField` that stores the instance's ID, the original question, the original
+       passage text, both of these in tokenized form, and the gold answer strings, accessible as
+       `metadata['id']`, `metadata['question']`, `metadata['context']`, `metadata['question_tokens']`,
+       `metadata['context_tokens']`, and `metadata['answers']`. This is so that we can more easily use the
        official SQuAD evaluation script to get metrics.
+     * `cls_index`, a `IndexField` that holds the index of the `[CLS]` token within the
+       `question_with_context` field. This is only present if `use_cls_token_for_unanswerable`
+       is set to `True`.
```
dirkgr, line 43: Can't we just find that token? Do we have to specify it in the instance?

allennlp_models/rc/dataset_readers/transformer_squad.py
```diff
@@ -50,11 +66,19 @@ class TransformerSquadReader(DatasetReader):
         is called "stride" instead of "overlap" because that's what it's called in the original huggingface
         implementation.
     skip_invalid_examples: `bool`, optional (default=`False`)
-        If this is true, we will skip examples that don't have a gold answer. You should set this to True during
-        training, and False any other time.
+        If this is true, we will skip examples where the question+context is truncated according to `length_limit`
+        such that the context no longer contains a gold answer. For SQuAD v1.1-style datasets, you should set this
+        to `True` during training, and `False` any other time.
```
dirkgr, line 71: What should I do for 2.2-style datasets?

allennlp_models/rc/dataset_readers/transformer_squad.py
```diff
@@ -50,11 +66,19 @@ class TransformerSquadReader(DatasetReader):
...
     max_query_length : `int`, optional (default=`64`)
         The maximum number of wordpieces dedicated to the question. If the question is longer than this, it will be
         truncated.
+    use_cls_token_for_unanswerable : `bool`, optional (default=`False`)
+        If this is true, we'll use the `[CLS]` token as the gold answer for questions that don't have a gold
+        answer. For SQuAD v2.0-style datasets that include impossible questions, this should
+        be set to `True`.
```
dirkgr, line 78: When would I set this to `False`?

allennlp_models/rc/dataset_readers/transformer_squad.py
```diff
@@ -50,11 +66,19 @@ class TransformerSquadReader(DatasetReader):
...
+    cls_token : `str`, optional (default=`[CLS]`)
+        The `[CLS]` token of the tokenizer. If `use_cls_token_for_unanswerable` is `False`, you don't
+        need to worry about this.
```
dirkgr, line 81: Can't we get this from the tokenizer? Why do we have to specify it here?

allennlp_models/rc/dataset_readers/transformer_squad.py
```diff
@@ -250,7 +294,10 @@ def text_to_instance(
...
                 question_field,
             )
-        else:
+        elif for_training and self.use_cls_token_for_unanswerable:
+            cls_index = fields["cls_index"].sequence_index
+            fields["answer_span"] = SpanField(cls_index, cls_index, question_field)
+        elif for_training:
             # We have to put in something even when we don't have an answer, so that this instance can be batched
             # together with other instances that have answers.
             fields["answer_span"] = SpanField(-1, -1, question_field)
```
dirkgr, line 303: Can't we do this all the time, and not have the extra flag at all?

allennlp_models/rc/dataset_readers/transformer_squad.py
```diff
@@ -222,9 +257,10 @@ def text_to_instance(
         tokenized_question: List[Token],
         context: str,
         tokenized_context: List[Token],
-        answers: List[str],
-        token_answer_span: Optional[Tuple[int, int]],
+        answers: List[str] = None,
+        token_answer_span: Optional[Tuple[int, int]] = None,
         additional_metadata: Dict[str, Any] = None,
+        for_training: bool = False,
```
dirkgr, line 263: I'd rather have flag names that explain what they do, instead of referring to the context to set them in.

For example, this flag could be called `always_add_answer_span`.

allennlp_models/rc/models/transformer_qa.py
```diff
@@ -65,131 +78,187 @@ def forward(  # type: ignore
...
-            loss = (start_loss + end_loss) / 2
+    def _evaluate_span(
+        self,
+        best_spans: torch.Tensor,
+        span_start_logits: torch.Tensor,
+        span_end_logits: torch.Tensor,
+        answer_span: torch.Tensor,
+    ) -> torch.Tensor:
+        """
+        Calculate the loss against the `answer_span` and also update the  span metrics.
```
dirkgr, line 216: double space

allennlp_models/rc/models/transformer_qa.py
```diff
@@ -65,131 +78,187 @@ def forward(  # type: ignore
...
+            zip(metadata, best_spans, context_span)
+        ):
+            context_tokens_for_question = metadata_entry["context_tokens"]
+
+            if cls_index is not None and best_span[0] == cls_index[i]:
+                # Predicting [CLS] is interpreted as predicting the question as unanswerable.
+                best_span_string = ""
+                # NOTE: even though we've "detached" 'best_spans' above, this still
+                # modifies the original tensor in-place.
+                best_span[0], best_span[1] = -1, -1
```
dirkgr, line 260: Do we want that to happen?

dirkgr (changes requested):
This looks great by and large.

The one thing I don't understand is why we can't just treat all questions as potentially unanswerable, and thus get rid of a bunch of flags and special cases. When we train on 1.1, the model would learn that questions are never unanswerable, and on 2.0 it would learn that sometimes they are. It could still be the same model.

allennlp_models/modelcards/rc-transformer-qa.json
epwalsh: Oh darn, I uploaded it with the wrong date. Meant to put "10" not "08"

epwalsh:
> The one thing I don't understand is why we can't just treat all questions as potentially unanswerable, and thus get rid of a bunch of flags and special cases. When we train on 1.1, the model would learn that questions are never unanswerable, and on 2.0 it would learn that sometimes they are. It could still be the same model.

That's true, I just wanted to be careful to not break backwards compatibility, at least not until our 2.0 release.

allennlp_models/rc/models/transformer_qa.py
epwalsh: Yes because `best_spans` (from each `best_span`) are added to the output dict. And it's expected that this span is either a valid index into the context, or something else. And I think the "something else" makes sense to be `(-1, -1)`.

allennlp_models/rc/dataset_readers/transformer_squad.py
epwalsh: I don't think there's a way to get this from the tokenizer 100% of the time, because some tokenizers, like RoBERTa, don't have a `[CLS]` token, right?

allennlp_models/rc/dataset_readers/transformer_squad.py
epwalsh: I don't know of a reliable way to find it

allennlp_models/rc/dataset_readers/transformer_squad.py
epwalsh: If we did this all the time, the model would end up computing a loss even when you're just trying to get predictions.

allennlp_models/rc/dataset_readers/transformer_squad.py
dirkgr: You can read it out of the huggingface tokenizers. They have calls for that.

allennlp_models/rc/dataset_readers/transformer_squad.py
dirkgr: It's just called `<s>` in RoBERTa.

allennlp_models/rc/dataset_readers/transformer_squad.py
dirkgr: Is that a problem?

It would not create a loss when we just don't have an answer.

allennlp_models/rc/models/transformer_qa.py
dirkgr: Modifying the input parameters is not usually the way we return information. We can break the rules in some cases, but when we do we should slap a big warning on the function that this is what's happening.

allennlp_models/rc/dataset_readers/transformer_squad.py
epwalsh: `answer_span` is the answer though.

epwalsh:
Ok, this is ready for another review. The dataset reader and model now infer all of the `[CLS]` token / index information so it doesn't need to be specified by the user.

I also trained the model again on RoBERTa large and got very good results (F1 score of 0.89).

allennlp_models/modelcards/rc-transformer-qa.json
```diff
@@ -37,8 +38,7 @@
         "model_performance_measures": "F1-score, Span Accuracy, Exact Match. Note that the metrics 
                                        that the model produces are calculated on a per-instance basis only.
                                        Since there could be more than one instance per question, these
-                                       metrics are not the official numbers on the SQuAD task. To get
-                                       official numbers, run the script in scripts/transformer_qa_eval.py.",
```
dirkgr, line 41: Did the script not survive?

allennlp_models/rc/dataset_readers/transformer_squad.py
```diff
@@ -20,41 +20,68 @@
...
-        If this is true, we will skip examples that don't have a gold answer. You should set this to True during
-        training, and False any other time.
+        If this is true, we will skip examples where the question+context is truncated according to `length_limit`
+        such that the context no longer contains a gold answer.
+
+        For SQuAD v1.1-style datasets, you should set this to `True` during training, and `False` any other time.
+
+        For SQuAD v2.0-style datasets, leaving this as `False` (recommended) will treat invalid examples
+        the same way as impossible examples, i.e. by setting the answer span to the span of the `[CLS]`
+        token.
```
dirkgr, line 79: Maybe we should call it "skip impossible instances"? I think SQuAD 2.0 uses the word "impossible" to identify instances without answers.

test_fixtures/rc/transformer_qa/experiment_v2.jsonnet
```diff
@@ -0,0 +1,31 @@
+local transformer_model = "albert-base-v2";
```
dirkgr, line 1: Didn't you make a special nano-BERT for these cases?

training_config/rc/transformer_qa.jsonnet
```diff
@@ -1,38 +1,40 @@
...
-      //"max_instances": 200  // debug setting
+    "type": "transformer_squad",
+    "transformer_model_name": transformer_model,
+    "length_limit": length_limit,
+    // "max_instances": 1000,  // debug setting
   },
-  "validation_dataset_reader": self.dataset_reader + {
-      "skip_invalid_examples": false,
+  "train_data_path": "https://allennlp.s3.amazonaws.com/datasets/squad/squad-train-v2.0.json",
+  "validation_data_path": "https://allennlp.s3.amazonaws.com/datasets/squad/squad-dev-v2.0.json",
```
dirkgr, line 17: How does this pass the test if we can't substitute the inputs?

dirkgr (approved):
Looks great, only minor concerns and questions!

training_config/rc/transformer_qa.jsonnet
epwalsh: `max_instances` is overridden and set to a small number in the tests, so the substitution is really only necessary for datasets that are not public.

test_fixtures/rc/transformer_qa/experiment_v2.jsonnet
epwalsh: Yes, I'll give that a try

allennlp_models/modelcards/rc-transformer-qa.json
epwalsh: It was moved, but it only works for SQuAD 1.1

PR #137, opened 2020-09-26
Adding more documentation to model cards spec

AkshitaB:


allennlp_models/common/model_card.py
```diff
@@ -42,6 +58,60 @@ def __str__(self):
...
+        The date on which the model was contributed. This is useful for all
+        stakeholders to become further informed on what techniques and
+        data sources were likely to be available during model development.
+        Format example: 2020-09-23
+
+    version: `str`
+        The version of the model, and how it differs from previous versions.
+        This is useful for all stakeholders to track whether the model is the
+        latest version, associate known bugs to the correct model versions,
+        and aid in model comparisons.
```
dirkgr, line 91: I would just expect a number here, or a number with dots. It might even be nice if these were comparable, so you can find out which one is more recent programmatically. I can't square that with writing down "how it differs from previous versions". How do you intend for this to be used?

allennlp_models/common/model_card.py
```diff
@@ -42,6 +58,60 @@ def __str__(self):
...
+        Eg. Naive Bayes Classifier.
+
+    paper: `str`
+        The paper on which the model is based.
+        Format example:
+        [Model Cards for Model Reporting (Mitchell et al, 2019)]
+        (https://api.semanticscholar.org/CorpusID:52946140)
+
+    citation: `str`
+        The BibTex for the paper.
```
dirkgr, line 107: Is it useful to have that in here? Semantic Scholar already has a link for that.

dirkgr (approved):

allennlp_models/common/model_card.py
AkshitaB: I think the intention for "how it differs from previous versions" is something like a changelog. Eg. "v1.2; we retrained the model using dataset-2.0, which fixes bug x observed in v1.1".

PR #139, opened 2020-09-30
fix seq2seq reader bug

epwalsh:
Addresses https://github.com/allenai/allennlp/issues/4687

### DIFF
--- a/allennlp_models/generation/dataset_readers/seq2seq.py
+++ b/allennlp_models/generation/dataset_readers/seq2seq.py
@@ -11,7 +11,7 @@ from allennlp.common.util import START_SYMBOL, END_SYMBOL
 from allennlp.data.dataset_readers.dataset_reader import DatasetReader
 from allennlp.data.fields import TextField
 from allennlp.data.instance import Instance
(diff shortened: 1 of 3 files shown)


AkshitaB (approved):

allennlp_models/generation/dataset_readers/seq2seq.py
```diff
@@ -78,13 +78,29 @@ def __init__(
...
+        self._end_token: Optional[Token] = None
+        if (
+            source_add_start_token
+            or source_add_end_token
+            or target_add_start_token
+            or target_add_end_token
+        ):
+            try:
+                self._start_token, self._end_token = self._source_tokenizer.tokenize(
+                    start_symbol + " " + end_symbol
```
fursovia, line 96: @epwalsh this will not work for character tokenizer. (even if `len(start_symbol) == 1` and `len(end_symbol) == 1`, because there is a space which is also a token)

any ideas how to fix this?

epwalsh: In that case you could have your tokenizer add any special symbols while disabling this behavior in the dataset reader with `source_add_start_token=False`.

PR #143, opened 2020-10-07
update masked lm language model

epwalsh:
Just needed to tweak the config. Weights are the same.


epwalsh:
I'm going to go ahead and merge this because it's such a small change and it's blocking me on something else.

PR #145, opened 2020-10-09
fix TransformerQA predictor test

epwalsh:
Quick fix to an oversight I missed in the SQuAD updates. This test randomly fails sometimes. This fixes that.


dirkgr (commented):
Can we just replace the example with something that shouldn't fail?

dirkgr:
Oh, it's running an untrained model. So no amount of sense in the example will make this behave properly.

epwalsh:
Right

dirkgr (approved):
This is kind of a screwy test now, but Akshita is depending on this being merged right now, so let's merge it and fix it if it becomes a problem later.

PR #159, opened 2020-10-29
Use standard util function in masked language model instead of hack

matt-gardner:
This fixes a bug in hotflip usage.


epwalsh (approved):
Current CI failure is fixed with this: https://github.com/allenai/allennlp-models/pull/158

PR #160, opened 2020-11-02
Updating pair classification model cards

AkshitaB:
Note: It appears that the ESIM model is fairly old (as per the training config). The archive file currently listed in the model card is not for the ESIM model, and it does not appear anywhere on our demo.

### DIFF
--- a/allennlp_models/modelcards/pair-classification-esim.json
+++ b/allennlp_models/modelcards/pair-classification-esim.json
@@ -3,7 +3,7 @@
     "registered_model_name": "esim",
     "registered_predictor_name": "textual_entailment",
     "display_name": "Enhanced LSTM for Natural Language Inference",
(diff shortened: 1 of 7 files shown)


allennlp_models/modelcards/pair-classification-esim.json
```diff
@@ -1,6 +1,59 @@
 {
     "id": "pair-classification-esim",
     "registered_model_name": "esim",
+    "registered_predictor_name": "textual_entailment",
     "display_name": "Enhanced LSTM for Natural Language Inference",
-    "archive_file": "decomposable-attention-elmo-2020.04.09.tar.gz"
+    "archive_file": "decomposable-attention-elmo-2020.04.09.tar.gz", //FIX: this file is the same as one for decomposable...
```
epwalsh, line 6: I think this might be the correct archive file:

```suggestion
    "archive_file": "esim-elmo-2020.02.10.tar.gz"
```

epwalsh (approved):

PR #161, opened 2020-11-02
Task cards

AkshitaB:
Motivation: A model's abilities and limitations come from the underlying task and the dataset it is trained on, in addition to the model itself. Therefore, it is helpful to separate out the information in model cards into 3 different concepts:

Task Card: Issues with defining the task itself. Eg. Span selection as a substitute for reading comprehension.
Data Card/Sheet: Biases in the data, etc. Eg. imbalanced demographics.
Model Card (Model specific): Architecture, algorithmic choices; loss function and metric choices. Eg. differences in picking L2 (average) error and L1 (median) error.

This PR adds a `TaskCard` class and task cards for some tasks in our demo. An additional advantage of this abstraction is that it can be directly used in the demo.


allennlp_models/common/task_card.py
```diff
@@ -0,0 +1,62 @@
...
+    """
+
+    id: str
+    name: Optional[str] = None
+    description: Optional[str] = None
+    expected_inputs: Optional[str] = None
+    expected_outputs: Optional[str] = None
+    scope_and_limitations: Optional[str] = None
+    examples: Optional[Union[List[Dict[str, str]], Dict[str, List[Dict[str, str]]]]] = None
+    # TODO: Add this to model cards.
```
epwalsh, line 62: So will `ModeCard` have a field for `task`?

AkshitaB: Yes, a reference to `task_id`.

epwalsh: Can that get added in this PR?

tests/pretrained_test.py
```diff
@@ -407,3 +408,7 @@ def test_transformer_qa(self):
     def test_pretrained_models(self, model_id, model_card):
         # Each model in pretrained_models should have an archive and registered_predictor_name.
         assert model_card.archive_file is not None
+
+    @pytest.mark.parametrize("task_id, task_card", get_tasks().items())
```
epwalsh, line 412: I think we should also have a test that ensures the `task_id` for each model card is valid.

epwalsh (approved):

PR #162, opened 2020-11-03
make naqanet test less flaky

epwalsh:
The gradient checks in this test are flaky as they involve checking that the gradient of a couple of bias tensors (of size 1) are non-zero.


dirkgr (commented):
Why is it OK for the bias to receive no gradients?

epwalsh:
> Why is it OK for the bias to receive no gradients?

I *think* there is a fairly high chance that the gradients of these small tensors end up close enough to zero that our checks consider them zero. I did this hastily though. There's probably a better, more general solution to this.

epwalsh:
I've seen this same issue with other models as well (it's always with small bias tensors like these).

dirkgr:
@matt-gardner, is that a thing? Small bias tensors not receiving gradients?

epwalsh:
Just updated to set the seed instead of explicitly ignoring the bias gradients.

dirkgr:
I like that better as a solution, but I'm still wondering if the bias thing has merit. If that is a common problem, it would be good to have your fix for it. I've just never heard of it.

epwalsh:
Latest failure comes from the same issue with a different model.

dirkgr (approved):

epwalsh:
I was able to reproduce this with a simple `torch.nn.Linear` model. It's just an issue of precision (or lack of).

So one thing we could do is just add a special case to `ModelTestCase.check_gradients` that will ignore scalar bias tensors with zero grads.

dirkgr:
If you're sure it's a precision issue, let's go with your original solution. No need to wait for @matt-gardner to weigh in.

PR #163, opened 2020-11-04
Adds a test that checks the interpret functionality

dirkgr:
I made this to check on a user's problem, so I might as well check it in.

Edit: Somehow this ended up being a bunch of random changes to make tests pass again. I suspect they were broken in `master` the whole time.

### DIFF
--- a/tests/classification/interpret/sst_test.py
+++ b/tests/classification/interpret/sst_test.py
@@ -1,0 +1,21 @@
+import pytest
+
+
+def test_gradient_visualization():
+    from allennlp.predictors.predictor import Predictor
+
+    predictor = Predictor.from_path(
+        "https://storage.googleapis.com/allennlp-public-models/sst-roberta-large-2020.06.08.tar.gz"
+    )
+    sentence = "a very well-made, funny and entertaining picture."
+
+    inputs = {"sentence": sentence}
+    from allennlp.interpret.saliency_interpreters import SimpleGradient
+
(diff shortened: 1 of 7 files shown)


dirkgr:
Python 3.8 🤛!

dirkgr:
It was actually down to the version of the `transformers` library. It has nothing to do with Python 3.8.

tests/classification/interpret/sst_test.py
```diff
@@ -0,0 +1,21 @@
...
+    inputs = {"sentence": sentence}
+    from allennlp.interpret.saliency_interpreters import SimpleGradient
+
+    simple_gradient_interpreter = SimpleGradient(predictor)
+    simple_gradient_interpretation = simple_gradient_interpreter.saliency_interpret_from_json(
+        inputs
+    )
+
+    gradients = simple_gradient_interpretation["instance_1"]["grad_input_1"]
+    assert max(gradients) - min(gradients) < 0.75
```
AkshitaB, line 21: Why do we check this difference? And where do we get `0.75` from?

dirkgr: When this was broken, the output was all `0` except for one cell that was `1`. I just want to check whether there is some variety in the numbers. Since they all add up to `1`, this seemed like a good check. It's possible that checking for just the `max` would be good enough.

AkshitaB (approved):

PR #165, opened 2020-11-12
allow for range of AllenNLP versions

epwalsh:
So I got to thinking that the core library tends to have a lot more activity than the models repo, and therefore it's likely that we'd want to do patch releases in the core library without bothering to do a corresponding patch release of models.

But we can't do that right now since every models release is pinned to the exact patch version of AllenNLP. For example, if we wanted to release AllenNLP 1.2.2 without doing a models release of 1.2.2, someone with models version 1.2.1 installed would have a dependency conflict if they tried to upgrade the core library to 1.2.2.

This PR proposes that we slightly decouple models from core, by allowing for a range of versions of the core library. In particular, if models is at 1.2.1, its core dependency would be `allennlp>=1.2.1,<1.3` instead of `allennlpl==1.2.1`. So this forces us to do a corresponding modes releases for every major or minor core release, but not for core patch releases.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -54,8 +54,7 @@ jobs:
 
     - name: Set AllenNLP version override
       run: |
-        VERSION=$(./scripts/get_version.py current --minimal)
-        ALLENNLP_VERSION_OVERRIDE="allennlp==$VERSION"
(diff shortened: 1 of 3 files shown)


schmmd:
How much work does it save to avoid a patch release of models?

epwalsh:
It would save waiting time. You wouldn't have to wait around for the core release to pass through CI in order to trigger the models release.

dirkgr (approved):

PR #166, opened 2020-11-14
Update model cards

AkshitaB:
* Added `short_description` for a one-line summary for the model.
* Miscellaneous information updates for all model cards.

### DIFF
--- a/tests/pretrained_test.py
+++ b/tests/pretrained_test.py
@@ -406,8 +406,11 @@ class TestAllenNlpPretrainedModels(AllenNlpTestCase):
 
     @pytest.mark.parametrize("model_id, model_card", get_pretrained_models().items())
     def test_pretrained_models(self, model_id, model_card):
-        # Each model in pretrained_models should have an archive and registered_predictor_name.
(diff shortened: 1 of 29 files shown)


tests/pretrained_test.py
```diff
@@ -406,8 +406,11 @@ def test_transformer_qa(self):
 
     @pytest.mark.parametrize("model_id, model_card", get_pretrained_models().items())
     def test_pretrained_models(self, model_id, model_card):
-        # Each model in pretrained_models should have an archive and registered_predictor_name.
+        # Each model in pretrained_models should have an archive and a description.
         assert model_card.archive_file is not None
+        assert model_card.display_name is not None
```
jonborchardt, line 411: we should add id to this list

jonborchardt (approved):
looks fine to me, but someone else should take a look at the actual values....

epwalsh (approved):
LGTM!

tests/pretrained_test.py
AkshitaB: A model card cannot be instantiated without `id`. So, no need to check it again.

PR #167, opened 2020-11-16
Fix bad start or end symbol error

JohnGiorgi:
This is a small PR that fixes what _I think_ was a typo in the error message of a `ValueError` thrown by `Seq2SeqDatasetReader`. Instead of printing out the "bad" `start_symbol` and `end_symbol`, it was printing the literal strings `"start_symbol"` and `"end_symbol"`. This wasn't caught by the unit tests, so I also updated the relevant test to catch it.

I ran into this because I actually triggered the error, and realized it wasn't reporting to me the "bad" symbols I was using.


epwalsh (approved):
Good catch! Thanks!

PR #169, opened 2020-11-17
More explicit start/end symbol parsing

JohnGiorgi:
Replaces the current check in `Seq2SeqDatasetReader` with a more explicit check that doesn't fail in the case of BPE-based tokenizers. This is based on a discussion [here](https://github.com/allenai/allennlp/issues/4798).

### DIFF
--- a/allennlp_models/generation/dataset_readers/seq2seq.py
+++ b/allennlp_models/generation/dataset_readers/seq2seq.py
@@ -46,9 +46,15 @@ class Seq2SeqDatasetReader(DatasetReader):
         Indexers used to define output (target side) token representations. Defaults to
         `source_token_indexers`.
     source_add_start_token : `bool`, (optional, default=`True`)
-        Whether or not to add `START_SYMBOL` to the beginning of the source sequence.
+        Whether or not to add `start_symbol` to the beginning of the source sequence.
     source_add_end_token : `bool`, (optional, default=`True`)
-        Whether or not to add `END_SYMBOL` to the end of the source sequence.
+        Whether or not to add `end_symbol` to the end of the source sequence.
+    start_symbol : `str`, (optional, default=`START_SYMBOL`)
+        The special token to add to the beginning of the target sequence, and the source sequence
+        if `source_add_start_token`.
+    end_symbol : `str`, (optional, default=`END_SYMBOL`)
+        The special token to add to the end of the target sequence, and the source sequence if
+        `source_add_end_token`.
     delimiter : `str`, (optional, default=`"\t"`)
         Set delimiter for tsv/csv file.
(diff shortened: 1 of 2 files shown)


allennlp_models/generation/dataset_readers/seq2seq.py
```diff
@@ -91,15 +97,17 @@ def __init__(
...
         ):
-            try:
-                self._start_token, self._end_token = self._source_tokenizer.tokenize(
-                    start_symbol + " " + end_symbol
-                )
-            except ValueError:
+            # Check that the tokenizer correctly appends the start and end tokens to
+            # the sequence without splitting them.
+            tokens = self._source_tokenizer.tokenize(start_symbol + " " + end_symbol)
+            start_token, end_token = tokens[0], tokens[-1]
```
epwalsh, line 103: We should also catch the potential `IndexError` that could happen here.

epwalsh (changes requested):

allennlp_models/generation/dataset_readers/seq2seq.py
JohnGiorgi: Okay, I wrapped it in a try/except. It's a little ugly, let me know if you see a cleaner way to check.

allennlp_models/generation/dataset_readers/seq2seq.py
```diff
@@ -91,16 +101,22 @@ def __init__(
...
+            # the sequence without splitting them.
+            tokens = self._source_tokenizer.tokenize(start_symbol + " " + end_symbol)
             try:
-                self._start_token, self._end_token = self._source_tokenizer.tokenize(
-                    start_symbol + " " + end_symbol
-                )
-            except ValueError:
+                start_token, end_token = tokens[0], tokens[-1]
+                if start_token.text != start_symbol or end_token.text != end_symbol:
+                    raise IndexError
```
epwalsh, line 110: My only issue with this that it will produce a stack trace that is a little confusing.

How about something like this:

```python
err_msg = f"Bad start or end symbol ('{start_symbol}', '{end_symbol}') for tokenizer {self._source_tokenizer}"
try:
    start_token, end_token = tokens[0], tokens[-1]
except IndexError:
    raise ValueError(err_msg)
if start_token.text != start_symbol or end_token.text != end_symbol:
    raise ValueError(err_msg)
```

JohnGiorgi: Yup, good call this is better.

epwalsh (approved):
LGTM!  Thanks @JohnGiorgi!

JohnGiorgi:
Thanks for the help :)

PR #174, opened 2020-11-23
Fix BART bug in prediction

wlhgtc:
The original `make_output_human_readable ` return same prediction(`predictions[0]`) in one batch.


wlhgtc:
By the way, I wonder why we need 2 special tokens in [task_step](https://github.com/allenai/allennlp-models/blob/master/allennlp_models/generation/models/bart.py#L340) ?

wlhgtc:
@epwalsh

PR #178, opened 2020-12-03
Treat auxiliary verbs as verbs

wangrat:
Hi, 
while working with this package I stumbled upon an issue where open information extraction predictor (`Predictor.from_path("https://storage.googleapis.com/allennlp-public-models/openie-model.2020.03.26.tar.gz")`) was giving me very satisfying results for complicated sentences, but for the ones that looked very simple it didn't return any results. At least this is how the issue looked like to me initially, with the examples I used.
Example of a sentence that didn't work: 
> Albert Einstein was a physicist.

I dig into the library to discover what was going on. I found out that in that sentence `was` isn't tagged as a `VERB` by Spacy. Spacy tags it as `AUX`, which stands for auxiliary verb. (https://universaldependencies.org/docs/u/pos/)

Because of that no instance is generated by `_sentence_to_srl_instances` and no results come up in the end.

I did a very simple fix of changing `if word.pos_ == "VERB":` to `if word.pos_ in ["VERB", "AUX"]:` and it solved the problem.

Now the question is, is it a bug, or a feature? If it's a bug then I'm glad I could help, if it's a feature then sorry for wasting your time.


wangrat:
Ah, sorry, I will update the changelog :)

wangrat:
I see that it now fails `TestSrlPredictor.test_prediction_with_no_verbs` test because now it recognises `have` as a verb.
But I don't understand why the test requires it not to recognise `have`. Is it not a verb?

With my change, you get this and you can extract some information from this.
```
{'verbs': [{'verb': 'seems',
   'description': '[ARG1: This sentence] [V: seems] [C-ARG1: to have a verb] .',
   'tags': ['B-ARG1',
    'I-ARG1',
    'B-V',
    'B-C-ARG1',
    'I-C-ARG1',
    'I-C-ARG1',
    'I-C-ARG1',
    'O']},
  {'verb': 'have',
   'description': 'This sentence seems to [V: have] [ARG1: a verb] .',
   'tags': ['O', 'O', 'O', 'O', 'B-V', 'B-ARG1', 'I-ARG1', 'O']}],
 'words': ['This', 'sentence', 'seems', 'to', 'have', 'a', 'verb', '.']}
```
Without my change, you get this and you have nothing. Is it really the desired behaviour?
```
{
            "verbs": [
                {
                    "verb": "seems",
                    "description": "This sentence seems to have a verb .",
                    "tags": ["O", "O", "O", "O", "O", "O", "O", "O"],
                }
            ],
            "words": ["This", "sentence", "seems", "to", "have", "a", "verb", "."],
        }
```

schmmd:
> Now the question is, is it a bug, or a feature? If it's a bug then I'm glad I could help, if it's a feature then sorry for wasting your time.

@wangrat that's for the PR--it's definitely a help!  I think this behavior changed when we upgraded SpaCy at some point, and we think this change is a good fix that we should merge.

dirkgr:
These tests fail when we change how the model works. If the output of the model makes sense, it's fine to change the test to reflect the new kind of model output we're getting. In fact, I'm happy that this happens, because it means that the change has test coverage.

wangrat:
I think we need to come to a final conclusion in https://github.com/allenai/allennlp/issues/4818 thread.
Then I'm happy to adjust my changes if necessary and fix/change the tests.

wangrat:
Alright, @AkshitaB, ready for review.

AkshitaB (approved):
Looks good to me. Thank you for the PR!

KTRosenberg:
@AkshitaB That's great to hear! When can we expect this to be available via pip? --or should I just build from source?

schmmd:
@KTRosenberg we will do a release at some point, but until then you can always install from a particular commit using pip as well.  For example, see https://stackoverflow.com/questions/13685920/install-specific-git-commit-with-pip.

dirkgr:
We are probably not releasing 1.3 this year, so in the meantime, I recommend cloning the repo, and installing with `pip install -e .`.

KTRosenberg:
That makes sense. Is "have" working now too? That one test with all "O"s suggest that have still isn't a verb, which I find confusing.

schmmd:
I'm not expert at the code, but I think that test case is demonstrating that now "have" is identified as a verb whereas before it wasn't.

KTRosenberg:
Got it. Thanks again for working on this. I really appreciate it.

There was one other thing I was wondering about though--is there any reason that AllenNLP's SRL doesn't do sense/roles and frames? (e.g. run.01). I noticed that [this functionality sort of exists](https://github.com/allenai/allennlp-models/blob/ef5d4229f62f3ea8f44345d43b6a7fd1ab2d09fa/allennlp_models/common/ontonotes.py#L255), but maybe the model is really out-of-date. What is the story with this?

Tangent over.

Thanks again.

wangrat:
@KTRosenberg I was aiming to fix the issue with using `Predictor.from_path("https://storage.googleapis.com/allennlp-public-models/openie-model.2020.03.26.tar.gz")` model as I mentioned in the PR description. 

That is openIE model which also uses the SrlPredictor class. It solved the issue for that model and for my use case - "have" is not "O" anymore.

I'm afraid your problem with srl model might be maybe somewhere deeper, like in the model itself, but I don't know honestly. In my case the model knew how to predict sensible information from "is" and "have" and similar, but the SRLPredictor using Spacy's POS prediction unnecessarily filtered them out before model could even have a go at them.

KTRosenberg:
> @KTRosenberg I was aiming to fix the issue with using `Predictor.from_path("https://storage.googleapis.com/allennlp-public-models/openie-model.2020.03.26.tar.gz")` model as I mentioned in the PR description.
> 
> That is openIE model which also uses the SrlPredictor class. It solved the issue for that model and for my use case - "have" is not "O" anymore.
> 
> I'm afraid your problem with srl model might be maybe somewhere deeper, like in the model itself, but I don't know honestly. In my case the model knew how to predict sensible information from "is" and "have" and similar, but the SRLPredictor using Spacy's POS prediction unnecessarily filtered them out before model could even have a go at them.

Ah, I see. In your opininion, are the results from that model competitive with the BERT model? Have you seen any particular trade-offs? I happen to be using Spacy for other things, so if the OpenIE model doesn’t match pos with Spacy, that might be a problem. I wonder if I could run both models and still get fast results, and match them. (I’d like to take this conversation offline if possible. I’m working on a project in which I could use aome guidence of using these different models.

I haven’t tried the pr yet, so it might work. I just noticed one of the tests feom the pr seemed to exclude “have” deliberately, but maybe I misinterpreted.

wangrat:
I'm not an expert on that, sorry. I tested just now `https://storage.googleapis.com/allennlp-public-models/bert-base-srl-2020.11.19.tar.gz` and `https://storage.googleapis.com/allennlp-public-models/openie-model.2020.03.26.tar.gz` with my fix and they both work for "have" and "was". And for the sentences I came up with they returned exactly same output.

I think I will be comparing these two later in my project, so far I'm prototyping and I don't really know which one I will end up using.

KTRosenberg:
That’s fair. If the openie one has extra role info, that’s useful.

Anyway, thanks!

wangrat:
Here is an example from tests you were referring to. Tests must be using some other model. In the screenshot `srl_predictor` is the srl model I mentioned in my previous comment and `rel_predictor` is the openIE model I mentioned.

Screenshot 2020-12-10 at 16 15 12

And here you can actually see a difference between these two models.

KTRosenberg:
Oh, I thouhht the openie model output verb role information. Is that something that must be enabled?

schmmd:
@KTRosenberg the Open IE task does not include role information, so there is no option to enable this.

KTRosenberg:
@schmmd Are there any models for SRL that are high-quality and contain role information? I couldn't find a way to browse which models are available on the google cloud api pages.

wangrat:
@KTRosenberg https://github.com/allenai/allennlp-models/tree/master/allennlp_models/modelcards I think these are the available models.

PR #184, opened 2020-12-16
fix cache volume

epwalsh:
See https://github.com/allenai/allennlp/pull/4869 and https://github.com/allenai/allennlp-demo/pull/659


epwalsh:
Merging so other runs don't mess up the cache.

PR #191, opened 2021-01-10
Fix small bugs in RC model

gladuo:


epwalsh (approved):
These changes LGTM, but how did the fix the dead lock? Or what was the reason for the dead lock in the first place?

epwalsh:
Also please run `black .` to fix the formatting issues when you get a chance

gladuo:
> These changes LGTM, but how did this fix the dead lock? Or what was the reason for the dead lock in the first place?
@epwalsh 
emmm... I might confuse you, but those changes above did NOT fix the dead lock, and they only fixed some other critical bugs in rc models.
I have not figured out what on earth made the dead lock, and I just bypassed it using drop_last by now.

epwalsh:
Gotcha, could you please fix the formatting so we can get this merged?

gladuo:
`black .` Done.

CHANGELOG could be like...

MOD correct a_gold and a_pred positions … a6cfeb9
ADD datatype for tensor … 1fc9a15
ADD is_training to make predictor work … 126b7e0

PR #193, opened 2021-01-15
Fix TransformerQA distributed dead-lock bug

epwalsh:
Closes https://github.com/allenai/allennlp/issues/4886

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - 'master' branch renamed to 'main'
 
+### Fixed
+
+- Fixed the potentially for a dead-lock when training the `TransformerQA` model on multiple GPUs
+  when nodes receive different sized batches.
+
 ## [v1.3.0](https://github.com/allenai/allennlp-models/releases/tag/v1.2.2) - 2020-12-15
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - 'master' branch renamed to 'main'
+- `SquadEmAndF1` metric can now also accept a batch of predictions and corresponding answers (instead of a single one)
+  in the form of list (for each).
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Fixed the potentially for a dead-lock when training the `TransformerQA` model on multiple GPUs
+- Fixed the potential for a dead-lock when training the `TransformerQA` model on multiple GPUs
   when nodes receive different sized batches.
 
 ## [v1.3.0](https://github.com/allenai/allennlp-models/releases/tag/v1.2.2) - 2020-12-15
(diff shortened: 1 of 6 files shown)


allennlp_models/rc/models/transformer_qa.py
```diff
@@ -247,9 +249,30 @@ def _collect_best_span_strings(
...
+        if is_distributed():
+            # In the distributed case, if the batch isn't balanced across nodes, we'll get a dead-lock
+            # if one node has more instances in a batch than another node.
+            if dist.get_backend() == "nccl":
+                device = torch.cuda.current_device()
+            else:
+                device = torch.device("cpu")
+            _min_node_batch_size = torch.tensor(min_node_batch_size, dtype=torch.int, device=device)
+            dist.all_reduce(_min_node_batch_size, op=dist.ReduceOp.MIN)
+            min_node_batch_size = _min_node_batch_size.item()
```
epwalsh, line 262: This is becoming a really common pattern. It would be to have a helper function for this called `dist_reduce` or something. Then we could just do:

```suggestion
            min_node_batch_size = dist_reduce(min_node_batch_size, dist.ReduceOp.MIN, dtype=torch.int)
```

epwalsh:
I don't think there's an easy way to write a test for this.

allennlp_models/rc/models/transformer_qa.py
dirkgr: Sounds like a good change.

dirkgr (commented):
There is no way to make this work correctly with uneven batch sizes, so that it doesn't throw anything away?

epwalsh:
There might be a way. I'll investigate some more tomorrow.

epwalsh:
@dirkgr updated. Now can handle different batch sizes.

allennlp_models/rc/metrics/squad_em_and_f1.py
```diff
@@ -23,15 +23,32 @@ def __init__(self) -> None:
...
+        self,
+        best_span_string: Union[str, List[str]],
+        answer_strings: Union[List[str], List[List[str]]],
+    ):
+        if not isinstance(best_span_string, list):
+            best_span_string = [best_span_string]
+            answer_strings = [answer_strings]  # type: ignore
+
+        cast(List[str], best_span_string)
+        cast(List[List[str]], answer_strings)
```
dirkgr, line 36: I never knew you could do that.

allennlp_models/rc/metrics/squad_em_and_f1.py
```diff
@@ -23,15 +23,32 @@ def __init__(self) -> None:
...
-    def __call__(self, best_span_string, answer_strings):
-        exact_match = squad.metric_max_over_ground_truths(
-            squad.compute_exact, best_span_string, answer_strings
-        )
-        f1_score = squad.metric_max_over_ground_truths(
-            squad.compute_f1, best_span_string, answer_strings
-        )
+    def __call__(
+        self,
+        best_span_string: Union[str, List[str]],
```
dirkgr, line 28: Should be called `best_span_strings` now?

CHANGELOG.md
```diff
@@ -14,6 +14,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - 'master' branch renamed to 'main'
+- `SquadEmAndF1` metric can now also accept a batch of predictions and corresponding answers (instead of a single one)
+  in the form of list (for each).
+
+### Fixed
+
+- Fixed the potentially for a dead-lock when training the `TransformerQA` model on multiple GPUs
```
dirkgr, line 22: "potential"

dirkgr (approved):
This is great! If it comes up with the other metrics, we should use this as a template for how to fix the others.

PR #195, opened 2021-01-19
Fix BART

epwalsh:
Getting good results now:

```python
# tmp.py
from allennlp.data import Instance, Vocabulary, Batch
from allennlp.data.fields import TextField
from allennlp.data.tokenizers import PretrainedTransformerTokenizer
from allennlp.data.token_indexers import PretrainedTransformerIndexer
from allennlp_models.generation.models import Bart
from allennlp.nn.util import move_to_device

model_name = "facebook/bart-large-cnn"
ARTICLE_TO_SUMMARIZE = """
The tower is 324 metres (1,063 ft) tall, about the same height as an 81-storey building,
and the tallest structure in Paris.
Its base is square, measuring 125 metres (410 ft) on each side.
During its construction, the Eiffel Tower surpassed the Washington Monument to
become the tallest man-made
structure in the world, a title it held for 41 years until the Chrysler Building in
New York City was finished in 1930.
It was the first structure to reach a height of 300 metres.
Due to the addition of a broadcasting aerial at the top of the tower in 1957,
it is now taller than the Chrysler Building by 5.2 metres (17 ft).
Excluding transmitters, the Eiffel Tower is the second tallest free-standing structure in
France after the Millau Viaduct.
""".strip().replace("\n", " ")

tokenizer = PretrainedTransformerTokenizer(model_name)
token_indexers = {"tokens": PretrainedTransformerIndexer(model_name)}
vocab = Vocabulary.empty()
model = Bart(model_name, vocab, indexer=token_indexers["tokens"]).cuda().eval()


def predict(text: str):
    tokens = tokenizer.tokenize(text)
    instance = Instance({"source_tokens": TextField(tokens, token_indexers)})
    instance.index_fields(vocab)
    batch = Batch([instance])
    tensor_dict = move_to_device(batch.as_tensor_dict(), 0)
    result = model(**tensor_dict)
    string = (
        "".join([t.text for t in result["predicted_tokens"][0]])
        .replace("Ġ", " ")
        .replace("<s>", "")
        .replace("</s>", "")
    )
    return string


print(predict(ARTICLE_TO_SUMMARIZE))
```

```
$> python tmp.py
The tower is 324 metres (1,063 ft) tall, about the same height as an 81-storey building.
Its base is square, measuring 125 metres (410 ft) on each side.
```

### DIFF
--- a/allennlp_models/generation/models/bart.py
+++ b/allennlp_models/generation/models/bart.py
@@ -365,7 +365,7 @@ class Bart(Model):
         predicted_tokens = [None] * predictions.shape[0]
         for i in range(predictions.shape[0]):
             predicted_tokens[i] = self._indexer.indices_to_tokens(
-                {"token_ids": [idx for idx in predictions[i].tolist() if idx != self._end_id]},
+                {"token_ids": predictions[i].tolist()},
                 self.vocab,
             )
         output_dict["predicted_tokens"] = predicted_tokens  # type: ignore
--- a/allennlp_models/generation/models/bart.py
+++ b/allennlp_models/generation/models/bart.py
@@ -216,7 +216,7 @@ class Bart(Model):
                 attention_mask=input_mask,
                 decoder_input_ids=target_ids[:, :-1].contiguous(),
                 decoder_attention_mask=target_mask[:, :-1].contiguous(),
-                use_cache=not self.training,
+                use_cache=False,
                 return_dict=True,
             )
             outputs["decoder_logits"] = bart_outputs.logits
@@ -309,8 +309,6 @@ class Bart(Model):
             A tuple containing logits for the next tokens of shape `(group_size, target_vocab_size)` and
             an updated state dictionary.
         """
-        print("Step:", step)
-
         if len(last_predictions.shape) == 1:
             last_predictions = last_predictions.unsqueeze(-1)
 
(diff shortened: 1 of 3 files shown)


epwalsh:
Tests will fail until https://github.com/allenai/allennlp/pull/4872 is merged.

allennlp_models/generation/models/bart.py
```diff
@@ -263,29 +264,28 @@ def forward(
...
     @staticmethod
-    def _decoder_cache_to_dict(decoder_cache):
+    def _decoder_cache_to_dict(decoder_cache: DecoderCacheType) -> Dict[str, torch.Tensor]:
         cache_dict = {}
         for layer_index, layer_cache in enumerate(decoder_cache):
-            for attention_name, attention_cache in layer_cache.items():
-                for tensor_name, cache_value in attention_cache.items():
-                    key = (layer_index, attention_name, tensor_name)
-                    cache_dict[key] = cache_value
+            assert len(layer_cache) == 4
```
dirkgr, line 270: Why are there four layers? I assume this is related to the internal huggingface changes?

dirkgr (approved):
Thanks for tackling this!

I assume we need to re-train the pre-trained model?

allennlp_models/generation/models/bart.py
epwalsh: Yea, this is internal to HF. Each layer cache contains the key and value tensors for the self-attention and cross-attention. So 4 tensors per layer. I'll add a comment on that.

epwalsh:
> I assume we need to re-train the pre-trained model?

These changes only affect validation / prediction, so I don't think it's necessary to retrain. What model are you talking about though? Did we actually fine-tune BART on something? It might be worth re-training since the validation metric was probably off.

PR #196, opened 2021-01-20
Interpret

leo-liuzy:
- fix a related allennlp issue: detailed in https://github.com/allenai/allennlp/issues/4745
- add the option in SNLI dataset reader to collapse "contradiction" and "neutral" into "non-entailment"


allennlp_models/pair_classification/dataset_readers/snli.py
```diff
@@ -15,6 +15,19 @@
 logger = logging.getLogger(__name__)
 
 
+def maybe_collapse_label(label: str, collapse: bool):
```
epwalsh, line 18: These changes should probably be in a separate PR

PR #205, opened 2021-01-29
VQA model card

AkshitaB:
* Added VQA model card.
* Renamed VE model card.


jonborchardt (approved):
What about the pretrained model id? See slack for details.

PR #206, opened 2021-01-29
OpenIE fix

AkshitaB:
Fixes https://github.com/allenai/allennlp/issues/4943
* Fixed `OpenIePredictor.predict_json` so it treats auxiliary verbs as verbs when the language is English.


epwalsh (approved):
LGTM!

PR #217, opened 2021-02-16
BART fix special characters in generated output

vikigenius:
Fixes https://github.com/allenai/allennlp/issues/4977

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,9 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `coding_scheme` parameter is now deprecated in `Conll2000DatasetReader`, please use `convert_to_coding_scheme` instead.
 
-### Fixed
+### Added
 
-- Fixed BART model so that it no longer produces special characters in `make_output_human_readable`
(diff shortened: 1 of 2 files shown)


epwalsh (changes requested):
Sorry that I didn't say this before, but I think we should keep `predicted_tokens` as is so that we don't break backwards compatibility. I think it's fine to have both `predicted_tokens` and `predicted_text`.

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `coding_scheme` parameter is now deprecated in `Conll2000DatasetReader`, please use `convert_to_coding_scheme` instead.
 
+### Fixed
+
+- Fixed BART model so that it no longer produces special characters in `make_output_human_readable`
```
epwalsh, line 16:
```suggestion
### Added

- BART model now adds a `predicted_text` field in `make_output_human_readable` that has the cleaned text corresponding to `predicted_tokens`.
```

epwalsh (approved):
LGTM!

PR #222, opened 2021-02-23
Fix prediction for models using TransformerMCReader

mkardas:
Hi, I stumbled upon an issue when playing with QA models, f.e.,:
```python
from allennlp_models.pretrained import load_predictor
predictor = load_predictor("mc-roberta-commonsenseqa")

question = "If I am tilting a drink toward my face, what should I do before the liquid spills over?"
alternatives = ["open mouth", "eat first", "use glass"]
predictor.predict(question, alternatives)
```

```python
(...)
     20     def _json_to_instance(self, json_dict: JsonDict) -> Instance:
     21         return self._dataset_reader.text_to_instance(
---> 22             "no_qid", json_dict["prefix"], json_dict["alternatives"]
     23         )

TypeError: text_to_instance() missing 1 required positional argument: 'label'
```
This PR fixes the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - BART model now adds a `predicted_text` field in `make_output_human_readable` that has the cleaned text corresponding to `predicted_tokens`.
 
+### Fixed
+
+- Made `label` parameter in `TransformerMCReader.text_to_instance` optional with default of `None`.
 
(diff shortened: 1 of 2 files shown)


epwalsh:
Thanks for catching this @mkardas! Could you just update the `CHANGELOG` with something like:

```
### Fixed

- Made `label` parameter in `TransformerMCReader.text_to_instance` optional with default of `None`.
```

mkardas:
Thanks @epwalsh, I've updated the changelog.

epwalsh (approved):
LGTM!

PR #226, opened 2021-03-11
Added model card for LERC and task card for evaluating-RC

anthonywchen:
Hello! I'm trying to add a new demo to AI2 for this [paper](https://arxiv.org/abs/2010.03636). 

The demo will add LERC, a metric for evaluating reading comprehension. In order to add to the demo, I've add to this repo:
- a new model card for LERC
- a new task card for the task of evaluating reading comprehension

Please let me know if there's anything else that would need to be added to this repository before adding to `allennlp-demo`.

Thanks!


anthonywchen:
Tagging @dirkgr as a follow up from our email conversations. Thanks!

allennlp_models/modelcards/evaluate_rc-lerc.json
```diff
@@ -0,0 +1,68 @@
...
+    "model_caveats_and_recommendations": {
+        "caveats_and_recommendations": null
+    },
+    "model_ethical_considerations": {
+        "ethical_considerations": null
+    },
+    "model_usage": {
+        "archive_file": "lerc-2020-11-18.tar.gz",
+        "training_config": null,
+        "install_instructions": "pip install allennlp==1.0.0 allennlp-models==1.0.0"
```
dirkgr, line 66: Does your system work with AllenNLP 2.1?

allennlp_models/taskcards/evaluate_rc.json
```diff
@@ -0,0 +1,34 @@
...
...ers\": null,\n        \"out_of_scope_use_cases\": null\n    },\n    \"factors\": {\n        \"relevant_factors\": null,\n        \"evaluation_factors\": null\n    },\n    \"metrics\": {\n        \"model_performance_measures\": null,\n        \"decision_thresholds\": null,\n        \"variation_approaches\": null\n    },\n    \"evaluation_data\": {\n        \"dataset\": null,\n        \"motivation\": null,\n        \"preprocessing\": null\n    },\n    \"training_data\": {\n        \"dataset\": {\n            \"name\": \"DROP\",\n            \"url\": \"https://allennlp.org/drop\"\n        },\n        \"motivation\": null,\n        \"preprocessing\": null\n    },\n    \"quantitative_analyses\": {\n        \"unitary_results\": null,\n        \"intersectional_results\": null\n    },\n    \"model_caveats_and_recommendations\": {\n        \"caveats_and_recommendations\": null\n    },\n    \"model_ethical_considerations\": {\n        \"ethical_considerations\": null\n    },\n    \"model_usage\": {\n        \"archive_file\": \"drop-nmn-2020.04.04.tar.gz\",\n        \"training_config\": null,\n        \"install_instructions\": \"pip install allennlp==1.0.0 allennlp-models==1.0.0\"\n    }\n}",
```
dirkgr, line 6: I think something went wrong here.

dirkgr (changes requested):
Other than the copy and paste accident, this looks fine, but I thought you'd also add the model and training config and so on?

allennlp_models/modelcards/evaluate_rc-lerc.json
anthonywchen: I don't believe so. The reason is that this version I believe uses a newer version of transformers, and some of the keys changed slightly since the version that allennlp 1.0.0 used. 

I can retrain my model using allennlp 2.1 and with the updated transformer is needed?

allennlp_models/taskcards/evaluate_rc.json
anthonywchen: Thanks for catching this. will fix.

anthonywchen:
Hi Dirk, thanks for looking at the pull request! I'll make the appropriate changes. I also noticed that there was a check that didn't pass saying I needed to modify the CHANGELOG. Is this something I should do?

Regarding adding model and training config, I have a [private repository](github.com/anthonywchen/MOCHA) where that lives, similar to the NMN model in the demo. I didn't see NMN model/config files in allennlp-models and so I also didn't include it.

dirkgr:
Yes, please update the CHANGELOG. It's easy. One line of English text :-)

There are two possible scopes of the work:
1. What you have here. We add the model card and a demo. The rest of your system lives in your own repo. That's a fine way of doing it, and either way, a good first step.
2. Integrate your metric as a metric that people can choose in their evaluations. That means it has to run on 2.1, the model should live in allennlp-models, and we'd add some metric code to the core allennlp repo. This is a lot more work, but it could drive a lot of adoption. It's much easier to get someone to use your system (and cite you in their paper), when all they have to do is put the right string into their training configuration file.

It's up to you whether you want to do #2, either now or later.

anthonywchen:
Ah right! I do remember having this conversation about making LERC a Metric class. I really like this idea but let's do (1) since I think its a good first step. 

I'll update the CHANGELOG.

dirkgr (approved):

anthonywchen:
Hi Dirk, 

I see that there are some checks that haven't been completed yet. Is there anything else I need to do or just sit tight until the checks run?

dirkgr:
I’m not sure why they didn’t run, and I’m trying to find out. There is
nothing you need to do, it looks all good from your side. There is
something wrong with CI.

dirkgr:
I fixed some stuff. Now it wants you to put in a training config. I looked in https://github.com/anthonywchen/MOCHA/tree/main/lerc, and there are two training configs. I assume they have to be run in sequence to re-train the model? Is there any sense at all in putting in only one of them?

dirkgr:
We're reviewing the options for what you can put in there so that the demo renders the right page. This is the first model like this, so we don't have a template answer. Sorry for the delay.

anthonywchen:
Hi Dirk, 

No problem! Yes, there are two configs because there is a pre-training step followed by a fine-tuning step. 

Regarding training configs, I don't believe I need to add it into `allennlp-models` to get the demo to work. Nitish's NMN model lives in his own repo, and in his [Dockerfile](https://github.com/allenai/allennlp-demo/blob/master/api/allennlp_demo/nmn_drop/Dockerfile) in allennlp-demo, he pulls that repo. The model config doesn't live in `allennlp-models`.

In my own fork of `allennlp-demo`, I tested his Dockerfile, and was able to get the API working with LERC despite not having the config in `allennlp-models`. This is just my cursory understanding, but I believe that the only thing we need to add to `allennlp-models` is the model card to render onto the demo. The custom Dockerfile handles the rest.

dirkgr:
You are completely right about that. That entry in the model card is just about making sure the right thing shows up in the demo under "training".

anthonywchen:
Thanks! Yeah, my understanding is that check asking for the config should only be necessary if your model is defined within this repository.

dirkgr:
I have an answer about the training step. @jonborchardt and I think it's best if you follow this pattern:
```
"training_dataset": {
      "name": "Bla dataset",
      "notes": "To train this model follow these [instructions](http://url/bla).",
      "processed_url": null,
      "url": "https://www.bla/"
    },
```

This way it won't render the normal `allennlp train` command, but instead put up a link to your repo, which I think is exactly what we want.

anthonywchen:
Added a `note` field to the `training_data` section of the LERC model card.

anthonywchen:
Hi Dirk, is there anything I need to do to fix `Pretrained Models` test fail?

Best,
Anthony

dirkgr:
> does adding the notes field stop the allennlp train command from running?

No, all that does is change what gets displayed on the demo page.

dirkgr (approved):
Thank you!

anthonywchen:
Thank you very much for your help! :)

dirkgr:
What about next steps? Are we adding this to the demo?

anthonywchen:
Yup, I'll make a pull request into `allennlp-demo` shortly.

PR #228, opened 2021-03-15
add BoolQ dataset reader

wangyuxinwhy:
https://github.com/allenai/allennlp/issues/4996

SuperGLUE is a important benchmark in NLP, so I add BoolQ dataset reader to allennlp-models.


wangyuxinwhy:
This is my first PR, so I am not familiar with the whole process. If there is any problem, please let me know and I will correct it right away.

PS: in training_config/classification/boolq_roberta.jsonnet

The following three lines may need to be changed to the AllenNLP AWS URL

  "train_data_path": "data/train.jsonl",
  "validation_data_path": "data/val.jsonl",
  "test_data_path": "data/test.jsonl",

dirkgr:
@wangyuxinwhy, we are OK with pointing directly to the official place where these files get downloaded, but if that's not an option, I'm happy to upload the files to AWS (or Google Storage, actually). Can you either update the URLs to they point to the official source, or send me the files so I can put them on GSE?

dirkgr (commented):
Sorry for the delay. The code looks great! Can you post, maybe right here, what kind of scores it gets and how it compares to Facebook's official scores?

Other than that I suspect we have to get through some mechanical stuff like making all the test succeed, but that should be easy.

dirkgr:
Oh, looks like you've done everything already. All we need is a download path that works, and this should be ready to merge!

wangyuxinwhy:
@dirkgr , Thank you for your review, Now there are three things:

1. Download path. I suggest uploading it to AllenNLP's GSE. The download path of BOOLQ is [here](https://dl.fbaipublicfiles.com/glue/superglue/data/v2/BoolQ.zip) , and BoolQ is released under the Creative Commons Share-Alike 3.0 license, This is the part you need to work on. 😊
2. The Codecov target is 90%, but my test case does not meet the conditions, I will complete this part as soon as possible.
3. I will do a more detailed experiment on Boolq and Roberta, and post the experiment report here.

dirkgr:
The code coverage check is often wrong. Don't worry about it.

I'm getting `"best_validation_accuracy": 0.863914373088685`. Facebook reports 87.1%, but I think that's on the test set, right? I will merge this, but if you want to play with the hyper parameters some more and get those last 0.5%, I'd be happy to put in the updated parameters!

dirkgr (approved):

PR #229, opened 2021-03-18
Allow example categories to be ordered

dirkgr:
This is part of a multi-repo nightmare. We need
 * https://github.com/allenai/allennlp/pull/5059
 * https://github.com/allenai/allennlp-models/pull/229
 * https://github.com/allenai/allennlp-demo/pull/896

all at the same time.


dirkgr:
*This can't be merged as it is.* We need to merge allenai/allennlp#5059 first, then change the SHA that is PR points to before merging here. But it can be reviewed.

AkshitaB (approved):

PR #230, opened 2021-03-18
Add `ignore_loss_on_o_tags` to `CrfTagger.__init__()`

tarohi24:
Context: https://github.com/allenai/allennlp/issues/5058
While `CrfTagger.forward()` has the `ignore_loss_on_o_tags` option, 
there's no way to pass the option to `forward` unless its dataset reader does.
Since I don't think `DatasetReader` is responsible for handing training parameters,
I added the option into `__init__`.

### DIFF
--- a/allennlp_models/tagging/models/crf_tagger.py
+++ b/allennlp_models/tagging/models/crf_tagger.py
@@ -68,7 +68,7 @@ class CrfTagger(Model):
         {"tags": List, "score": float}.
         The "tags" value for the first dict in the list for each data_item will be the top
         choice, and will equal the corresponding item in output_dict['tags']
-    ignore_loss_on_o_tags : `bool`, optional (detaul=`False`)
+    ignore_loss_on_o_tags : `bool`, optional (default=`False`)
         If True, we compute the loss only for actual spans in `tags`, and not on `O` tokens.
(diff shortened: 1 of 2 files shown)


allennlp_models/tagging/models/crf_tagger.py
```diff
@@ -205,6 +212,7 @@ def forward(
         loss : `torch.FloatTensor`, optional
             A scalar loss to be optimised. Only computed if gold label `tags` are provided.
         """
+        ignore_loss_on_o_tags = ignore_loss_on_o_tags or self.ignore_loss_on_o_tags
```
epwalsh, line 215: I think this should be:

```suggestion
        ignore_loss_on_o_tags = ignore_loss_on_o_tags if ignore_loss_on_o_tags is not None else self.ignore_loss_on_o_tags
```

epwalsh (commented):
Hi @tarohi24, this is a good solution, thanks! I just have one comment 👇

allennlp_models/tagging/models/crf_tagger.py
tarohi24: Thanks @epwalsh for your review! I fixed it.

epwalsh (approved):

PR #231, opened 2021-03-23
Add copynext op to copynet

JohnGiorgi:
As discussed in https://github.com/allenai/allennlp/issues/5025, this PR adds a quasi [CopyNext](https://www.aclweb.org/anthology/2020.spnlp-1.2/) operation to CopyNet. This implementation simply allows the user to add a new `copy_next_token` to the target vocabulary, which when generated by the decoder leads to a copying of the previously copied token.

If the CopyNext operation is illegal at the current timestep, we will just literally decode the `copy_next_token`, e.g. `"@CN@"`.

@epwalsh Mind giving this a lookover? 

To test, I copied some of the existing tests from `CopyNetTest` and modified them accordingly. I also tried the code on my own task of joint NER/RE -- it works, but there is no performance gain.


epwalsh:
Thanks @JohnGiorgi!

I was thinking more about this the other day, and it occurred to me that this approach of just looking for the CN token probably doesn't help much with CopyNet (which you confirmed experimentally). I think the reason for this is that by using the CN token in place of the actual target token, you're loosing any information held by the actual target token (and any matching source tokens). And it also adds ambiguity, because the model could either copy a source token or generate the CN token, and in many cases this leads to the same target sequence.

I'm not sure if I'm explaining that well. It's hard to explain in a few sentences. But this got my thinking that there is probably a better way to add a inductive bias for copying spans into CopyNet. I'm not sure what that is yet, but I'm going to put some more thought into it.

JohnGiorgi:
> And it also adds ambiguity, because the model could either copy a source token or generate the CN token, and in many cases this leads to the same target sequence.

This is my intuition as well. 

Okay no problem! Let me know if you think of an implementation that makes more sense? I can also dwell on it.

If it makes more sense just to implement CopyNext from scratch, I could also work on that!

epwalsh:
Hey John, yea, if you want to implement the true CopyNext model, we'd be glad to accept that!

dirkgr:
What's the deal with this? Can we merge it?

dirkgr (commented):
Is it possible to add a training config that takes advantage of this, or will it take advantage automatically?

Either way, I'd love to have a comparison, before and after, though since this is not part of the demo, it's not strictly necessary.

epwalsh:
@dirkgr we are going to hold off until we find a better way to integrate this idea into CopyNet, or have a true implementation of CopyNext.

PR #232, opened 2021-03-24
Always keep Spacy PoS tagger on

leo-liuzy:
Fixing some fail tests in https://github.com/allenai/allennlp/pull/5066
Due to the different use of function `get_spacy_models`


leo-liuzy:
I found those changes are not necessary

PR #234, opened 2021-03-26
Make snli dataset reader work with unlabeled instances

nelson-liu:
`text_to_instance` supports the case where no label is provided, and this just modifies `_read` to handle unlabeled instances as well.


nelson-liu:
looks like one of the ci checks is stuck

AkshitaB (approved):

PR #235, opened 2021-03-28
Add make_output_human_readable to pair classification models

nelson-liu:
This makes it such that predictor output has a `label` key with `neutral`, `contradiction`, and `entailment`.


allennlp_models/pair_classification/models/bimpm.py
```diff
@@ -226,23 +225,33 @@ def add_matching_result(matcher, encoded_premise, encoded_hypothesis):
...
         """
-        Converts indices to string labels, and adds a `"label"` key to the result.
+        Does a simple argmax over the probabilities, converts index to string label, and
+        add `"label"` key to the dictionary with the result.
         """
-        predictions = output_dict["probs"].cpu().data.numpy()
-        argmax_indices = numpy.argmax(predictions, axis=-1)
-        labels = [self.vocab.get_token_from_index(x, namespace="labels") for x in argmax_indices]
-        output_dict["label"] = labels
+        predictions = output_dict["label_probs"]
```
nelson-liu, line 242: not sure what the right thing to do here is---bimpm outputs probs under the `probs` key, but the other models use `label_probs`. should we standardize bimpm and change `probs` to `label_probs` here, too?

dirkgr: Yes, I think that's the best change.

dirkgr (approved):

PR #238, opened 2021-03-30
Fixed bug in `experiment_from_huggingface.jsonnet` 

ArjunSubramonian:
Fixed bug in `experiment_from_huggingface.jsonnet` by changing `min_count` to have key `labels` instead of `answers`. Resolves failure of model checks that involve calling `_extend` in `vocabulary.py` (Related PR: https://github.com/allenai/allennlp/pull/3644)

PR #240, opened 2021-03-31
Fix label namespace in experiment.jsonnet

AkshitaB:
Related to https://github.com/allenai/allennlp-models/pull/238

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


ArjunSubramonian (commented):
Before I approve, I'm unsure if this fixes all the "answers"-related bugs, as there are other places where "answers" shows up in .jsonnet files. I'm having trouble reproducing the bug locally with the related allennlp changes.

AkshitaB:
@ArjunSubramonian I believe this is the only config that requires the fix, because that's the default namespace for the [VE model](https://github.com/allenai/allennlp-models/blob/main/allennlp_models/vision/models/visual_entailment.py#L52). For other configs that have `answers`, [the model](https://github.com/allenai/allennlp-models/blob/main/allennlp_models/vision/models/vilbert_vqa.py#L48) also has the namespace as `answers`.

ArjunSubramonian (approved):

PR #241, opened 2021-03-31
Add T5 for generation/summarization

epwalsh:


Makefile
```diff
@@ -20,8 +20,8 @@ DOCKER_RUN_CMD = docker run --rm \
 		-v $$HOME/.cache/huggingface:/root/.cache/huggingface \
 		-v $$HOME/nltk_data:/root/nltk_data
 
-# TODO: change this back to master branch
-ALLENNLP_COMMIT_SHA = $(shell git ls-remote https://github.com/allenai/allennlp main | cut -f 1)
+# TODO: change this back to main branch
+ALLENNLP_COMMIT_SHA = $(shell git ls-remote https://github.com/allenai/allennlp t5 | cut -f 1)
```
epwalsh, line 24:
```suggestion
ALLENNLP_COMMIT_SHA = $(shell git ls-remote https://github.com/allenai/allennlp main | cut -f 1)
```

requirements.txt
```diff
@@ -1,4 +1,4 @@
-git+https://github.com/allenai/allennlp.git@main
+git+https://github.com/allenai/allennlp.git@T5  # TODO: change back to @main
```
epwalsh, line 1:
```suggestion
git+https://github.com/allenai/allennlp.git@main
```

allennlp_models/generation/models/t5.py
```diff
@@ -0,0 +1,119 @@
...
+
+                for metric in self._metrics:
+                    metric(output_dict["predictions"], labels)  # type: ignore[call-arg]
+
+        return output_dict
+
+    @overrides
+    def make_output_human_readable(self, output_dict: Dict[str, torch.Tensor]) -> Dict[str, Any]:
+        predictions = output_dict["predictions"]
+        output_dict["predicted_text"] = self.tokenizer.tokenizer.batch_decode(
```
dirkgr, line 108: The normal way to do this is by looking into the vocab. Then you don't need a tokenizer in the model at all.

dirkgr (changes requested):
I think you can get away without having a tokenizer in the model.

allennlp_models/generation/models/t5.py
epwalsh: This does more than just convert IDs into tokens. It also de-tokenizes the tokens while removing any special tokens. Sure, we could re-write that logic here using just the vocab, but I don't think it's worth it.

dirkgr: Fair enough

dirkgr (approved):

PR #243, opened 2021-04-05
big improvements to readme

epwalsh:
While the situation is improving, model discoverability is still not great. This adds a lot of info to the README to help with that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,10 @@
 <div align="center">
     <br>
-    <img src="https://raw.githubusercontent.com/allenai/allennlp/main/docs/img/allennlp-logo-dark.png" width="400"/>
+    <a href="https://github.com/allenai/allennlp">
+      <img src="https://raw.githubusercontent.com/allenai/allennlp/main/docs/img/allennlp-logo-dark.png" width="400"/>
+    </a>
+    <br>
+    <br>
     <p>
     Officially supported AllenNLP models.
     </p>
--- a/README.md
+++ b/README.md
@@ -26,6 +26,100 @@
 ❗️ To file an issue, please open a ticket on <a href="https://github.com/allenai/allennlp/issues/new/choose">allenai/allennlp</a> and tag it with "Models". ❗️
 </div>
 
+## About
+
+This repository contains the components - such as [`DatasetReader`](https://docs.allennlp.org/main/api/data/dataset_readers/dataset_reader/#datasetreader), [`Model`](https://docs.allennlp.org/main/api/models/model/#model), and [`Predictor`](https://docs.allennlp.org/main/api/predictors/predictor/#predictor) classes - for applying [AllenNLP](https://github.com/allenai/allennlp) to a wide variety of NLP [tasks](#tasks-and-components).
+It also provides an easy way to download and use [pre-trained models](#pre-trained-models) that were trained with these components.
+
+### Tasks and components
+
+This is an overview of the tasks supported by the AllenNLP Models library along with the corresponding components provided, organized by category. For a more comprehensive overview, see the [AllenNLP Models documentation](https://docs.allennlp.org/models/main/) or the [Paperswithcode page](https://paperswithcode.com/lib/allennlp).
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -26,6 +26,100 @@
...
+## About
+
+This repository contains the components - such as [`DatasetReader`](https://docs.allennlp.org/main/api/data/dataset_readers/dataset_reader/#datasetreader), [`Model`](https://docs.allennlp.org/main/api/models/model/#model), and [`Predictor`](https://docs.allennlp.org/main/api/predictors/predictor/#predictor) classes - for applying [AllenNLP](https://github.com/allenai/allennlp) to a wide variety of NLP [tasks](#tasks-and-components).
+It also provides an easy way to download and use [pre-trained models](#pre-trained-models) that were trained with these components.
+
+### Tasks and components
+
+This is an overview of the tasks supported by the AllenNLP Models library along with the corresponding components provided, organized by category. For a more comprehensive overview, see the [AllenNLP Models documentation](https://docs.allennlp.org/models/main/) or the [Paperswithcode page](https://paperswithcode.com/lib/allennlp).
+
+- [**Classification**](https://github.com/allenai/allennlp-models/tree/main/allennlp_models/classification)
```
ArjunSubramonian, line 38: Perhaps add a description of what Classification is, like you've done for the other categories, e.g. prediction of discrete integer labels for examples

README.md
```diff
@@ -26,6 +26,100 @@
...
+  
+  Classification is a broad category that contains many more specific tasks such as Sentiment Analysis and Binary Question Answering.
+
+  🛠 **Components provided:** Dataset readers for various datasets, including [BoolQ](https://docs.allennlp.org/models/main/models/classification/dataset_readers/boolq/) and [SST](https://docs.allennlp.org/models/main/models/classification/dataset_readers/stanford_sentiment_tree_bank/), as well as a [Biattentive Classification Network](https://docs.allennlp.org/models/main/models/classification/models/biattentive_classification_network/) model.
+
+- [**Coreference Resolution**](https://github.com/allenai/allennlp-models/tree/main/allennlp_models/coref)
+
+  Coreference resolution is the task of finding all expressions that refer to the same entity in a text. It is an important step for many higher level NLP tasks that involve natural language understanding such as document summarization, question answering, and information extraction.
```
ArjunSubramonian, line 46: Should we cite https://nlp.stanford.edu/projects/coref.shtml here?

README.md
```diff
@@ -26,6 +26,100 @@
...
+
+- [**Language Modeling**](https://github.com/allenai/allennlp-models/tree/main/allennlp_models/lm)
+
+  Language Modeling tasks involve learning a probability distribution over sequences of tokens.
+
+  🛠 **Components provided:** Several language model implementations, such as a [Masked LM](https://docs.allennlp.org/models/main/models/lm/models/masked_language_model/) and a [Next Token LM](https://docs.allennlp.org/models/main/models/lm/models/next_token_lm/).
+
+- [**Multiple Choice**](https://github.com/allenai/allennlp-models/tree/main/allennlp_models/mc)
+
+  Multiple Choice tasks are tasks that involve selecting a correct answer out of a set of possible answers, given some input.
```
ArjunSubramonian, line 64: Personally confused on the distinction between `multiple choice` and `classification`

dirkgr (dismissed):
Big fan!

README.md
dirkgr: Multiple choice is choosing between alternatives. Classification is evaluating each alternative on its own. In practice we can get human-level performance by classifying each option on its own, but I suspect that's a problem with the datasets. I wonder if a model that can actually compare possible answers against each other could do even better than what we do now.

epwalsh: Just updated. Let me know if that makes sense.

README.md
epwalsh: Thanks, added!

README.md
epwalsh: Yeup, good call. I was being lazy 😜

ArjunSubramonian (approved):
Looks good to me!

PR #244, opened 2021-04-06
Add best span probs

jacob-morrison:
Adding span probabilities to TransformerQA's output to address [allenai/allennlp#4465](https://github.com/allenai/allennlp/issues/4465)

### DIFF
--- a/allennlp_models/rc/models/transformer_qa.py
+++ b/allennlp_models/rc/models/transformer_qa.py
@@ -185,10 +185,9 @@ class TransformerQA(Model):
         ) + torch.gather(span_end_logits, 1, best_spans[:, 1].unsqueeze(1))
         best_span_scores = best_span_scores.squeeze(1)
 
-        best_span_probs = (
-            torch.gather(span_start_probs, 1, best_spans[:, 0].unsqueeze(1))
-            + torch.gather(span_end_probs, 1, best_spans[:, 1].unsqueeze(1))
-        ) / 2
+        best_span_probs = torch.gather(
+            span_start_probs, 1, best_spans[:, 0].unsqueeze(1)
+        ) * torch.gather(span_end_probs, 1, best_spans[:, 1].unsqueeze(1))
         best_span_probs = best_span_probs.squeeze(1)
 
         output_dict = {
--- a/allennlp_models/rc/models/transformer_qa.py
+++ b/allennlp_models/rc/models/transformer_qa.py
@@ -186,9 +186,8 @@ class TransformerQA(Model):
         best_span_scores = best_span_scores.squeeze(1)
 
         best_span_probs = (
-            torch.gather(
-                span_start_probs, 1, best_spans[:, 0].unsqueeze(1)
-            ) + torch.gather(span_end_probs, 1, best_spans[:, 1].unsqueeze(1))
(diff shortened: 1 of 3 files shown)


dirkgr:
If you reference https://github.com/allenai/allennlp/issues/4465 in the description, merging this PR will auto-close the issue.

allennlp_models/rc/models/transformer_qa.py
```diff
@@ -178,10 +185,19 @@ def forward(  # type: ignore
         ) + torch.gather(span_end_logits, 1, best_spans[:, 1].unsqueeze(1))
         best_span_scores = best_span_scores.squeeze(1)
 
+        best_span_probs = (
+            torch.gather(span_start_probs, 1, best_spans[:, 0].unsqueeze(1))
+            + torch.gather(span_end_probs, 1, best_spans[:, 1].unsqueeze(1))
+        ) / 2
```
dirkgr, line 191: I don't think this is correct for probabilities. The probability of a span is the probability that it starts at the start token _times_ the probability that it ends at the end token.

tests/rc/models/transformer_qa_test.py
```diff
@@ -44,6 +44,10 @@ def test_forward_pass_runs_correctly(self):
         # script.
         assert metrics["per_instance_f1"] > 0
 
+        output_span_start_probs = output_dict["span_start_probs"][0].data.numpy()
+        output_span_end_probs = output_dict["span_end_probs"][0].data.numpy()
+        assert_almost_equal(numpy.sum(output_span_start_probs, -1), 1, decimal=6)
+        assert_almost_equal(numpy.sum(output_span_end_probs, -1), 1, decimal=6)
```
dirkgr, line 50: We probably can't do this style of test for `best_span_probs`, because we don't want to enumerate all possible spans, but we can test some basics like "They should all be greater than 0." and "Their sum must be <= 1.".

Though I'm looking at this through a cardboard tube view. If I'm missing something that makes this hard, don't bend over backwards to make it happen.

dirkgr (changes requested):

allennlp_models/rc/models/transformer_qa.py
jacob-morrison: Good catch! I was going off of what the person said in the issue, but that makes more sense to me.

tests/rc/models/transformer_qa_test.py
jacob-morrison: Done. It didn't seem too hard to add those checks, let me know if you think of anything else.

jacob-morrison:
> If you reference [allenai/allennlp#4465](https://github.com/allenai/allennlp/issues/4465) in the description, merging this PR will auto-close the issue.

Can it auto-close that issue if it's technically in a different repository?

dirkgr:
> Can it auto-close that issue if it's technically in a different repository?

Oh, right. I guess it cannot. Otherwise it would show up now on the right under "linked issues".

dirkgr (approved):

dirkgr:
Sweet, ship it!

dirkgr:
aba8e5e1555d2354a7233de495cb33e9

PR #246, opened 2021-04-07
Add list of pretrained models to README

epwalsh:
Also adds a script and a GitHub Actions job for keeping this list up-to-date. The script can be ran to update the list. The Actions job will fail if the script needs to be ran.

### DIFF
--- a/allennlp_models/modelcards/modelcard-template.json
+++ b/allennlp_models/modelcards/modelcard-template.json
@@ -8,7 +8,7 @@
         "archive_file": null,
         "training_config": null,
         "overrides": null,
-        "install_instructions": "pip install allennlp==1.0.0 allennlp-models==1.0.0",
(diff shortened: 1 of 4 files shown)


allennlp_models/modelcards/modelcard-template.json
```diff
@@ -8,7 +8,7 @@
         "archive_file": null,
         "training_config": null,
         "overrides": null,
-        "install_instructions": "pip install allennlp==1.0.0 allennlp-models==1.0.0",
+        "install_instructions": "pip install allennlp==1.0.0 allennlp-models==1.0.0"
```
dirkgr, line 11: The template should probably say `2.2.0` by now.

dirkgr (approved):
This reminds me, all the people working on models right now need to add a model card as well. Otherwise their work will disappear into obscurity!

PR #249, opened 2021-04-13
Making TransformerQAPredictor compatible with interpret modules

AkshitaB:
* Implements `predictions_to_labeled_instances` and `_json_to_instance`. 

Fixes https://github.com/allenai/allennlp-demo/issues/679

### DIFF
--- a/allennlp_models/rc/predictors/transformer_qa.py
+++ b/allennlp_models/rc/predictors/transformer_qa.py
@@ -93,8 +93,7 @@ class TransformerQAPredictor(Predictor):
         # if `qid` is `None`, it is updated using self._next_qid
         context = json_dict["passage"] if "passage" in json_dict else json_dict["context"]
         result: List[Instance] = []
-        if qid is None:
-            question_id = self._next_qid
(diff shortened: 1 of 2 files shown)


allennlp_models/rc/predictors/transformer_qa.py
```diff
@@ -47,23 +53,51 @@ def predict_json(self, inputs: JsonDict) -> JsonDict:
...
+        )
+        new_instance.add_field("answer_span", answer_span)
+        return [new_instance]
+
     @overrides
     def _json_to_instance(self, json_dict: JsonDict) -> Instance:
-        raise NotImplementedError(
-            "This predictor maps a question to multiple instances. "
-            "Please use _json_to_instances instead."
+        logger.warning(
```
AkshitaB, line 80: @dirkgr I'm not sure if just adding a warning instead of an error actually breaks anything.

dirkgr: I don't think it breaks anything, but people don't read warnings.

dirkgr (approved):

PR #250, opened 2021-04-14
Distributed VQA metric

dirkgr:


tests/vision/metrics/vqa_test.py
```diff
@@ -0,0 +1,127 @@
...
+            [-1, -1],
+            multiple_runs,
+            VqaMeasure(),
+            metric_kwargs,
+            desired_accuracy,
+            exact=True,
+        )
+
+
+def multiple_runs(
```
AkshitaB, line 110: Can we update the `global_distributed_metric` function to take the `number_of_runs` as a parameter instead of defining a new function? The default value could just be `1`.

AkshitaB (changes requested):

tests/vision/metrics/vqa_test.py
dirkgr: We can. It's been a loooong time since I had a change that affected only a single repo ...

dirkgr: https://github.com/allenai/allennlp/pull/5127

dirkgr:
😟 On my mac that test segfaults. I was hoping it was just a mac/python thing. But it looks like it's also happening on the test servers.

dirkgr:
This is the error:
```
terminate called after throwing an instance of 'gloo::EnforceNotMet'
  what():  [enforce fail at /opt/conda/conda-bld/pytorch_1616554788289/work/third_party/gloo/gloo/transport/tcp/pair.cc:490] op.preamble.length <= op.nbytes. 8 vs 4
```

dirkgr:
Looks like this happens when you try to sum tensors with different data types?

Why do we put up with APIs like that?

dirkgr:
@AkshitaB, this is ready for another review.

AkshitaB (approved):

PR #253, opened 2021-04-19
Remove registered_predictor_name from all modelcards where the predictor is the same as the default

AkshitaB:
Fixes https://github.com/allenai/allennlp/issues/5118

### DIFF
--- a/allennlp_models/modelcards/ve-vilbert.json
+++ b/allennlp_models/modelcards/ve-vilbert.json
@@ -1,7 +1,7 @@
 {
     "id": "ve-vilbert",
-    "registered_model_name": "visual-entailment",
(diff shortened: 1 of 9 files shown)


allennlp_models/modelcards/ve-vilbert.json
```diff
@@ -1,7 +1,7 @@
 {
     "id": "ve-vilbert",
-    "registered_model_name": "visual-entailment",
-    "registered_predictor_name": "vilbert_ve",
+    "registered_model_name": "ve_vilbert",
```
dirkgr, line 3: 👀

dirkgr (approved):
Are those the only cases? I thought this would be the case for virtually all the models?

PR #256, opened 2021-04-26
Update vilbert_vgqa_pretrained.jsonnet

epwalsh:


training_config/vision/vilbert_vgqa_pretrained.jsonnet
```diff
@@ -74,6 +74,9 @@ local vocabulary = if construct_vocab then {
   "data_loader": {
     "batch_size": gpu_batch_size,
     "shuffle": true,
+    "max_instances_in_memory": gpu_batch_size * 100,
```
ArjunSubramonian, line 77: For my understanding, where is the 100 coming from?

epwalsh: Totally arbitrary. In my benchmarking I found that `100 x batch_size` worked well.

PR #258, opened 2021-05-03
Add weighting option to CopyNet

epwalsh:
This is just a minor, general improvement to CopyNet. It adds the option to weight the loss contributions from individual instances when calculating the batch loss.

### DIFF
--- a/allennlp_models/generation/dataset_readers/copynet_seq2seq.py
+++ b/allennlp_models/generation/dataset_readers/copynet_seq2seq.py
@@ -157,8 +157,12 @@ class CopyNetDatasetReader(DatasetReader):
         # Parameters
 
         source_string : `str`, required
+
         target_string : `str`, optional (default = `None`)
+
         weight : `float`, optional (default = `None`)
+            An optional weight to assign to this instance when calculating the loss in
+            [CopyNetSeq2Seq.forward()](../../models/copynet_seq2seq/#forward.parameters).
 
         # Returns
 
@@ -199,7 +203,7 @@ class CopyNetDatasetReader(DatasetReader):
         fields_dict["metadata"] = MetadataField(meta_fields)
 
         if weight is not None:
-            fields_dict["weight"] = ArrayField(np.array([float(weight)]))
+            fields_dict["weight"] = ArrayField(np.array([float(weight)], dtype=np.single))
 
         return Instance(fields_dict)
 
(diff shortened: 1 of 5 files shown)


allennlp_models/generation/dataset_readers/copynet_seq2seq.py
```diff
@@ -194,6 +202,9 @@ def text_to_instance(
 
         fields_dict["metadata"] = MetadataField(meta_fields)
 
+        if weight is not None:
+            fields_dict["weight"] = ArrayField(np.array([float(weight)], dtype=np.single))
```
dirkgr, line 206: Just make it a `TensorField`, and use Torch methods to create the tensors. `ArrayField` is the old way.

allennlp_models/generation/models/copynet_seq2seq.py
```diff
@@ -527,7 +536,13 @@ def _forward_loss(
...
         log_likelihood = (log_likelihoods * target_mask).sum(dim=-1)
         # The loss is the negative log-likelihood, averaged over the batch.
-        loss = -log_likelihood.sum() / batch_size
+        if weight is None:
+            loss = -log_likelihood.sum() / batch_size
+        else:
+            # shape: (batch_size,)
+            if len(weight.shape) > 1:
+                weight = weight.squeeze()
+            loss = -(weight * log_likelihood).sum() / batch_size
```
dirkgr, line 545: Shouldn't this be `/ weight.sum()`?

dirkgr (changes requested):

allennlp_models/generation/models/copynet_seq2seq.py
epwalsh: I've gone back and forth on this.

If you use `weight.sum()` to normalize, then the weighting is only relative to each batch, which is probably *not* what you want.

For example, let's say we normalize by `weight.sum()` and your weights range from 0.5 - 1.0. If you have a batch that contains only instances with weights of 0.5, then this will give you the same result as if they all had weights of 1.0.

dirkgr: Hmm, fair enough. I would expect that setting all the weights to 1000 should be the same as setting them all to 0.001. But to get that behavior, and also the behavior you want, we would have to sum up all the weights in the dataset before processing a single batch, and then scale each batch accordingly. That's not practical. So let's leave it like this then.

dirkgr (approved):

PR #263, opened 2021-05-15
Improve transformer_squad reader to avoid duplicate tokenization of context in training

MagiaSN:
In training, the same context will be used in multiple instances, this pull request add cached_tokenized_context to avoid duplicate tokenization of the context. This reduce the preprocessing time from 30m49s to 13m35s on my machine, and yields exactly the same dev results as the original implementation.


epwalsh (approved):
This is a great improvement! Thanks @MagiaSN 🙂

PR #267, opened 2021-05-19
Specify BeamSearch as a parameter

JohnGiorgi:
Right now, the `BeamSearch` object in the generation models is mostly hardcoded, with only some of its arguments passed to a model constructor (e.g. `beam_size` and `max_steps`). Because `BeamSearch` has quite a few arguments that you might be interested in experimenting with (e.g. `min_steps`, `sampler`, `final_sequence_scorer`) it would be nice if the `BeamSearch` object could be constructed in the config file.

Unfortunately, that doesn't really work, as it would require a user to provide the `end_index` argument, which they (probably) don't know. <del>The next best solution (I think) is to update the generation models so that they accept a dictionary, `beam_search_kwargs`, that can then be passed to initialize a `BeamSearch` object, like so:

```python
self._beam_search = BeamSearch(self._end_index, **beam_search_kwargs)
```

<del>This is similar to the [`tokenizer_kwargs`](https://github.com/allenai/allennlp-models/blob/dea182c44b613e2d92d4b2a5c5515ae2d598f56b/allennlp_models/rc/dataset_readers/transformer_squad.py#L93) used elsewhere in AllenNLP.</del>

<del>As a little proof-of-concept I implemented this simple tweak for CopyNet and updated the tests. If you guys think this is useful, I can update the rest of the generation models and push them here!</del>

The new strategy, as discussed, is to

1. Make `BeamSearch` registrable, and register it with the name `"beam_search"` (https://github.com/allenai/allennlp/pull/5231)
2. Add a `Lazy[beam_search]` argument to the `generation` models.
3. Build the `BeamSearch` object lazily in the constructors of the `generation` models (with something `like self._beam_search = beam_search.construct(end_index=self._end_index)`


JohnGiorgi:
Hmm, I realized this doesn't work for the arguments that are custom classes, like `sampler` and `final_sequence_scorer`. E.g. if you try to pass something like

```jsonnet
{
    // Some args here
    "model": {
        // Some args here
        "beam_search_kwargs": {
            "final_sequence_scorer": {
                "type": "length-normalized-sequence-log-prob"
            }
        }
    }
}
```

`"final_sequence_scorer"` is interpreted as a `dict` and not constructed as a `FinalSequenceScorer` object. Not sure why this is.

dirkgr:
What's going on with this PR? I think the idea is a good one.

@JohnGiorgi, have you considered using `Lazy[]` for this? That way you could construct a `Lazy[BeamSearch]` object, and we give it the `end_index` when we call `.construct()` on the lazy object.

dirkgr:
@JohnGiorgi, an alternative approach would be what @danieldeutsch suggests at https://github.com/allenai/allennlp/issues/5219. I sort of prefer his suggestion because it keeps the API cleaner, but I want to make sure it addresses the problem you have.

JohnGiorgi:
@dirkgr Thanks for the feedback! I think both @danieldeutsch suggestion in https://github.com/allenai/allennlp/issues/5219 and your suggestion of using `Lazy[]` solve my problem! I basically just want a way to specify `BeamSearch` in the config.

I would be happy to implement @danieldeutsch solution here (or in another PR) but I also don't want to step on his toes so if he's already working on it I can close this!

dirkgr:
We talked about this case this morning. We think it makes the most sense to leave the `end_index` parameter in the constructor, but use `Lazy` to make beam search usable from a config file.

We'll talk in the other issue about moving `end_index`. I don't understand why that's necessary.

JohnGiorgi:
@dirkgr I think I understand how to use `Lazy[]` and `.construct()`. But I am not sure how to specify a `BeamSearch` object in the config as it [doesn't look like it is `Registrable`](https://github.com/allenai/allennlp/blob/3d5799d8784592d648fa2531bd26bb24e5831b0c/allennlp/nn/beam_search.py#L527).

Should I open a PR in AllenNLP to make it inherit from `Registrable` and give it a name, or is there another way around this?

dirkgr:
Yes, make it `Registrable`. Let's give it the name `"beam_search"`, and define it as the `default_implementation`. We'll almost never have to use the name anyways.

This will make the existing `BeamSearch` class the base class for all other potential beam search implementations, which is maybe a bit weird as an API. But it's backwards compatible, and we can address that issue later.

JohnGiorgi:
@dirkgr @AkshitaB Okay this is ready for review. It is going to depend on https://github.com/allenai/allennlp/pull/5231 being merged.

There are a few failing tests, I believe because they are loading model archives that contain configs that are now out of date. Not sure the best way to solve this.

allennlp_models/generation/models/bart.py
```diff
@@ -138,9 +137,8 @@ def __init__(
         self,
         model_name: str,
         vocab: Vocabulary,
+        beam_search: Lazy[BeamSearch],
```
epwalsh, line 140:
```suggestion
        beam_search: Lazy[BeamSearch] = Lazy(BeamSearch),
```

allennlp_models/generation/models/bart.py
```diff
@@ -126,10 +127,8 @@ class Bart(Model):
         Vocabulary containing source and target vocabularies.
     indexer : `PretrainedTransformerIndexer`, optional (default = `None`)
         Indexer to be used for converting decoded sequences of ids to to sequences of tokens.
-    max_decoding_steps : `int`, optional (default = `128`)
-        Number of decoding steps during beam search.
-    beam_size : `int`, optional (default = `4`)
-        Number of beams to use in beam search. The default is from the BART paper.
+    beam_search : `BeamSearch`, required
```
epwalsh, line 130:
```suggestion
    beam_search : `Lazy[BeamSearch]`, optional (default = `Lazy(BeamSearch)`)
```

allennlp_models/generation/models/copynet_seq2seq.py
```diff
@@ -46,10 +47,8 @@ class CopyNetSeq2Seq(Model):
     attention : `Attention`, required
         This is used to get a dynamic summary of encoder outputs at each timestep
         when producing the "generation" scores for the target vocab.
-    beam_size : `int`, required
-        Beam width to use for beam search prediction.
-    max_decoding_steps : `int`, required
-        Maximum sequence length of target predictions.
+    beam_search : `BeamSearch`, required
```
epwalsh, line 50:
```suggestion
    beam_search : `BeamSearch`, optional (default = `Lazy(BeamSearch)`)
```

allennlp_models/generation/models/copynet_seq2seq.py
```diff
@@ -76,8 +75,7 @@ def __init__(
         source_embedder: TextFieldEmbedder,
         encoder: Seq2SeqEncoder,
         attention: Attention,
-        beam_size: int,
-        max_decoding_steps: int,
```
epwalsh, line 80: It does make sense to remove these, but I'd like to keep backwards compatibility. So let's check for these parameters in `**kwargs`. If we find them in `**kwargs`, issue a `DeprecationWarning` and pass them through to `beam_search.construct()`.

allennlp_models/generation/models/simple_seq2seq.py
```diff
@@ -34,8 +35,8 @@ class SimpleSeq2Seq(Model):
         Embedder for source side sequences
     encoder : `Seq2SeqEncoder`, required
         The encoder of the "encoder/decoder" model
-    max_decoding_steps : `int`
-        Maximum length of decoded sequences.
```
epwalsh, line 38: Same comment about backwards compatibility.

allennlp_models/generation/models/simple_seq2seq.py
```diff
@@ -72,9 +71,8 @@ def __init__(
         vocab: Vocabulary,
         source_embedder: TextFieldEmbedder,
         encoder: Seq2SeqEncoder,
-        max_decoding_steps: int,
+        beam_search: Lazy[BeamSearch],
```
epwalsh, line 74:
```suggestion
        beam_search: Lazy[BeamSearch] = Lazy(BeamSearch),
```

allennlp_models/generation/models/simple_seq2seq.py
```diff
@@ -34,8 +35,8 @@ class SimpleSeq2Seq(Model):
         Embedder for source side sequences
     encoder : `Seq2SeqEncoder`, required
         The encoder of the "encoder/decoder" model
-    max_decoding_steps : `int`
-        Maximum length of decoded sequences.
+    beam_search : `BeamSearch`, required
```
epwalsh, line 38:
```suggestion
    beam_search : `BeamSearch`, optional (default = `Lazy(BeamSearch)`)
```

epwalsh (changes requested):
Just a few suggestions to keep backwards compatibility. The main thing is making the `beam_search` parameter optional, with a default value of `Lazy(BeamSearch)`.

allennlp_models/generation/models/copynet_seq2seq.py
JohnGiorgi: Good call, I agree. I addressed this in `copynet_seq2seq.py` and added a test in `test_copynet_seq2seq.py`. Mind taking a look? If you are happy with it I will do it for the rest of the models!

allennlp_models/generation/models/copynet_seq2seq.py
```diff
@@ -142,9 +140,19 @@ def __init__(
...
+            "The parameter {} as been deprecated."
+            "Provide this parameter as argument to beam_search instead."
         )
+        if "beam_size" in kwargs:
+            beam_search._params["beam_size"] = kwargs["beam_size"]
+            raise DeprecationWarning(deprecation_warning.format("beam_size"))
+        if "max_decoding_steps" in kwargs:
+            beam_search._params["max_steps"] = kwargs["max_decoding_steps"]
+            raise DeprecationWarning(deprecation_warning.format("max_decoding_steps"))
+        self._beam_search = beam_search.construct(end_index=self._end_index)
```
JohnGiorgi, line 155: @epwalsh Here is my solution for addressing your last outstanding comment about backwards compatibility (will do it for the rest of the models if you are happy with this).

tests/generation/models/copynet_test.py
```diff
@@ -20,6 +20,17 @@ def setup_method(self):
...
+        # These values are arbitrary but should be different than the config.
+        beam_size, max_decoding_steps = 100, 1000
+        params = Params.from_file(self.param_file)
+        params["model"]["beam_size"] = beam_size
+        params["model"]["max_decoding_steps"] = max_decoding_steps
+        with pytest.raises(DeprecationWarning):
+            model = Model.from_params(vocab=self.vocab, params=params.get("model"))
+            assert model._beam_search.beam_size == beam_size
+            assert model._beam_search.max_steps == max_decoding_steps
+
```
JohnGiorgi, line 33: @epwalsh andddd here's a test to make sure it actually works.

allennlp_models/generation/models/copynet_seq2seq.py
epwalsh: Hmm I'd rather not modify `beam_search._params` since it's a private attribute. Instead I would just make a `beam_search_extras = {}` variable and add `"beam_size"` and `"max_decoding_steps"` from `kwargs` into it, if found. Then do `beam_search.construct(end_index=self._end_index, **beam_search_extras)`.

JohnGiorgi: Right, good catch. Done! Shall I update the other generation models with this change?

epwalsh: Looks good! Go ahead with that and then we can get this merged ✅

JohnGiorgi: Awesome. Done. I had to tweak some of the tests under `generation`. Looks like AllenNLP treats `DeprecationWarning` as an error? From the `pytest.ini`:

"Our policy here is to ignore (silence) any deprecation warnings from _outside_ allennlp, but to treat any _internal_ deprecation warnings as errors."

That caused some trouble when a model config had `beam_size` or `max_decoding_steps`, as my fix catches those and then calls `DeprecationWarning`, causing the test to fail. In any case, all tests under `generation/models` and `generation/modules` are passing locally now.

epwalsh: Oh, interesting. I'm not sure when we decided that, but I'm going to disable that for now.

epwalsh: Ah, nevermind, I just updated those test fixtures

epwalsh (approved):
LGTM! Thanks for doing this @JohnGiorgi 🙂

JohnGiorgi:
My pleasure, thanks for the help as always!

PR #268, opened 2021-05-23
Added binary gender bias-mitigated RoBERTa model for SNLI

ArjunSubramonian:
Added binary gender bias-mitigated RoBERTa model for SNLI.

Depends on https://github.com/allenai/allennlp/pull/5176.


AkshitaB (commented):
Looks good, overall. Just the same question as in the other PR about needing the full `allennlp.fairness.bias_mitigator_applicator.BiasMitigatorApplicator` in the config.

ArjunSubramonian:
> Looks good, overall. Just the same question as in the other PR about needing the full `allennlp.fairness.bias_mitigator_applicator.BiasMitigatorApplicator` in the config.

In this case, yes, because `BiasMitigatorApplicator` is in `allennlp.fairness` and `FromParams` only looks at the modules imported in model's `__init__.py`. I tried placing `BiasMitigatorApplicator`  in model's `__init__.py`, but this caused a circular dependency issue for some reason - `BiasMitigatorApplicator` was using some fairness module classes that hadn't been imported yet, etc.

AkshitaB (approved):

ArjunSubramonian:
> > Looks good, overall. Just the same question as in the other PR about needing the full `allennlp.fairness.bias_mitigator_applicator.BiasMitigatorApplicator` in the config.
> 
> In this case, yes, because `BiasMitigatorApplicator` is in `allennlp.fairness` and `FromParams` only looks at the modules imported in model's `__init__.py`. I tried placing `BiasMitigatorApplicator` in model's `__init__.py`, but this caused a circular dependency issue for some reason - `BiasMitigatorApplicator` was using some fairness module classes that hadn't been imported yet, etc.

I'm just a little confused about why "snli" is not accepted as a registered `DatasetReader`. This is the reason why I have the full path of the reader specified. It seems fine for every other model, but I'm struggling to figure out what's causing an issue here.

tests/pair_classification/models/bias_mitigator_applicator_test.py
```diff
@@ -0,0 +1,47 @@
+from flaky import flaky
+import numpy
+
+from tests import FIXTURES_ROOT
+from allennlp.common.testing import ModelTestCase
+
```
AkshitaB, line 6:
```suggestion
from allennlp_models.pair_classification import SnliReader
from allennlp.fairness.bias_mitigator_applicator import BiasMitigatorApplicator
```

test_fixtures/pair_classification/bias_mitigation/experiment.json
```diff
@@ -0,0 +1,85 @@
...
+        "model_name": "epwalsh/bert-xsmall-dummy",
+        "max_length": 512
+      }
+    }
+  },
+  "train_data_path": "test_fixtures/pair_classification/bias_mitigation/snli_train.jsonl",
+  "validation_data_path": "test_fixtures/pair_classification/bias_mitigation/snli_dev.jsonl",
+  "test_data_path": "test_fixtures/pair_classification/bias_mitigation/snli_test.jsonl",
+  "model": {
+    "type": "allennlp.fairness.bias_mitigator_applicator.BiasMitigatorApplicator",
```
AkshitaB, line 21:
```suggestion
    "type": "bias_mitigator_applicator",
```

test_fixtures/pair_classification/bias_mitigation/experiment.json
```diff
@@ -0,0 +1,85 @@
+{
+  "dataset_reader": {
+    "type": "allennlp_models.pair_classification.dataset_readers.snli.SnliReader",
```
AkshitaB, line 3:
```suggestion
    "type": "snli",
```

AkshitaB (commented):
Those modules need to be imported in your test file. In the usual case, users would include the right package using `--include_package`.

PR #285, opened 2021-06-23
Flickr30k

jacob-morrison:
Adding a dataset reader, model, and training config for flickr30k image retrieval


allennlp_models/vision/models/vilbert_image_retrieval.py
```diff
@@ -0,0 +1,163 @@
...
+                rel_scores, indices = scores.topk(self.k, dim=-1)
+
+                # Shape: (batch_size)
+                pre_logits = torch.sum((indices == label.reshape(-1, 1)), dim=-1).float()
+
+                # Shape: (batch_size, 2)
+                # 1st column == 1 if we found the image in the top k, 2nd column == 1 if we didn't
+                logits = torch.stack((pre_logits, (pre_logits == 0).float()), dim=1)
+
+                outputs = {"logits": logits}
```
dirkgr, line 126: That's not what logits are. In this code, the "scores" would be the closest thing to logits? I don't know if that term makes sense here at all.

dirkgr: Did you call it that for compatibility with some other code?

dirkgr: Oh, I see. They are logits in the training code path, and it makes sense there.

dirkgr: The `CategoricalAccuracy` metric already supports top-k. Can you just use that instead of rolling your own?

allennlp_models/vision/models/vilbert_image_retrieval.py
```diff
@@ -0,0 +1,163 @@
...
+            label_namespace,
+            is_multilabel=False,
+            ignore_text=ignore_text,
+            ignore_image=ignore_image,
+        )
+        self.classifier = torch.nn.Linear(pooled_output_dim, 1)
+
+        self.accuracy = CategoricalAccuracy()
+        self.loss = CrossEntropyLoss()
+        self.fbeta = FBetaMeasure(beta=1.0, average="macro")
```
dirkgr, line 74: Do you use this anywhere?

allennlp_models/vision/models/vilbert_image_retrieval.py
```diff
@@ -0,0 +1,163 @@
...
+        self, output_dict: Dict[str, torch.Tensor]
+    ) -> Dict[str, torch.Tensor]:
+        batch_tokens = []
+        for batch_index, batch in enumerate(output_dict["probs"]):
+            tokens = {}
+            for i, prob in enumerate(batch):
+                tokens[self.vocab.get_token_from_index(i, self.label_namespace)] = float(prob)
+            batch_tokens.append(tokens)
+        output_dict["tokens"] = batch_tokens
+        return output_dict
```
dirkgr, line 161: This is copy&paste gore, isn't it? That looks like the function from VQA.

allennlp_models/vision/predictors/vilbert_ir.py
```diff
@@ -0,0 +1,36 @@
...
+            raise ValueError(
+                f"Dataset reader is of type f{self._dataset_reader.__class__.__name__}. "
+                f"Expected {Flickr30kReader.__name__}."
+            )
+
+    @overrides
+    def predictions_to_labeled_instances(
+        self, instance: Instance, outputs: Dict[str, numpy.ndarray]
+    ) -> List[Instance]:
+        return [instance]  # TODO
```
dirkgr, line 36: ?

dirkgr: If you don't have time to finish this, throw `NotImplementedError`.

dirkgr (changes requested):

allennlp_models/vision/predictors/vilbert_ir.py
jacob-morrison: This is copied from the vilbert_vqa predictor, but I think I've fixed it now either way.

allennlp_models/vision/models/vilbert_image_retrieval.py
jacob-morrison: Yup, I didn't see anything VQA-specific in it so I thought it looked reasonable, but let me know if I should change it.

allennlp_models/vision/models/vilbert_image_retrieval.py
jacob-morrison: No, I thought I deleted it. Thanks!

allennlp_models/vision/models/vilbert_image_retrieval.py
```diff
@@ -0,0 +1,163 @@
...
+            return outputs
+
+        else:
+            with torch.no_grad():
+                # Shape: (batch_size, num_images, pooled_output_dim)
+                self.backbone.eval()
+                pooled_output = self.backbone(box_features, box_coordinates, box_mask, caption)[
+                    "pooled_boxes_and_text"
+                ]
+                self.backbone.train()
```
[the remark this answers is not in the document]
jacob-morrison: Done.

allennlp_models/vision/dataset_readers/vision_reader.py
```diff
@@ -96,11 +96,13 @@ def __init__(
...
         write_to_cache: bool = True,
+        manual_distributed_sharding: bool = True,
+        manual_multiprocess_sharding: bool = True,
     ) -> None:
         super().__init__(
             max_instances=max_instances,
-            manual_distributed_sharding=True,
-            manual_multiprocess_sharding=True,
+            manual_distributed_sharding=manual_distributed_sharding,
+            manual_multiprocess_sharding=manual_multiprocess_sharding,
```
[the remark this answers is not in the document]
jacob-morrison: Yes, I got it to start training.

allennlp_models/vision/dataset_readers/vision_reader.py
```diff
@@ -232,6 +240,36 @@ def _coordinates_cache(self) -> MutableMapping[str, Tensor]:
...
+                self._hard_negative_coordinates_cache_instance = {}
+            else:
+                os.makedirs(self.feature_cache_dir, exist_ok=True)  # type: ignore
+                self._hard_negative_coordinates_cache_instance = TensorCache(
+                    os.path.join(self.feature_cache_dir, "hard_negative_coordinates"),  # type: ignore
+                    read_only=not self.write_to_cache,
+                )
+
+        return self._hard_negative_coordinates_cache_instance
+
```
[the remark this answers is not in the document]
jacob-morrison: Sure! I added it there because the eventual COCO IR reader will need the same stuff, but we can just move it back whenever we write that one.

allennlp_models/vision/dataset_readers/flickr30k.py
```diff
@@ -0,0 +1,469 @@
...
+    def text_to_instance(
+        self,
+        caption_dicts: List[Dict[str, Any]],
+        image_index: int,
+        caption_index: int,
+        features_list: List[Tensor] = [],
+        features_list_field: Optional[ListField] = None,
+        coordinates_list: List[TensorField] = [],
+        coordinates_list_field: Optional[ListField] = None,
+        masks_list: Optional[Tensor] = None,
```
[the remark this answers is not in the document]
jacob-morrison: We can (and we'd most likely pick the `List` version), but then we'll remake the ListField for each validation instance. So we'd make a ListField with size 1000 for each of the 5000 validation instances.

allennlp_models/vision/dataset_readers/flickr30k.py
```diff
@@ -0,0 +1,469 @@
...
+            masks.append(
+                ArrayField(
+                    hard_negative_features.new_ones(
+                        (hard_negative_features.shape[0],),
+                        dtype=torch.bool,
+                    ),
+                    padding_value=False,
+                    dtype=torch.bool,
+                )
+            )
```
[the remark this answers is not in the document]
jacob-morrison: There's no performance hit, but it wouldn't really simplify the signature. Either we pass in `hard_negative_features` and `hard_negative_coordinates`, or we pass in `averaged_features` and `caption_tensor` to calculate the hard negatives. I think the simplest way to slim down the method signature would be to find a way to pass in only one of each of the List/ListField parameter pairs.

allennlp_models/vision/models/vilbert_image_retrieval.py
```diff
@@ -0,0 +1,163 @@
...
+    def __init__(
+        self,
+        vocab: Vocabulary,
+        text_embeddings: TransformerEmbeddings,
+        image_embeddings: ImageFeatureEmbeddings,
+        encoder: BiModalEncoder,
+        pooled_output_dim: int,
+        fusion_method: str = "mul",
+        dropout: float = 0.1,
+        label_namespace: str = "answers",
```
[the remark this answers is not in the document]
jacob-morrison: I don't believe so, so I think we can remove this parameter.

allennlp_models/vision/dataset_readers/flickr30k.py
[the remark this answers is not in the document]
dirkgr: Making fields is not expensive. At least, making `ListField` is not expensive. It just keeps a reference to the actual data. So let's just pass in `List`.

allennlp_models/vision/dataset_readers/flickr30k.py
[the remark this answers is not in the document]
dirkgr: :+1:

allennlp_models/vision/models/vilbert_image_retrieval.py
dirkgr: This function assembles `output_dict["tokens"]`, because it thinks the answers it produces are tokens. This won't work in this model at all.

We can have this function do nothing. I don't know exactly how we would represent "the right image" in a `Dict` in a human-readable way. I guess if we had the original filename, that would be neat to add, but it's a bit of work to plumb that all the way through.

jacob-morrison: Oh I see, that makes sense. Thanks for clarifying that. I'll make it just return the output dictionary.

allennlp_models/vision/models/vilbert_image_retrieval.py
jacob-morrison: Hm. Will that work if I'm using the accuracy without top-k scoring for the training runs? Or should I create a second accuracy metric based on top-k scoring? And if I should create a second one, can you point me at a good example of how to use it? I'm not sure what its inputs should look like in top-k mode.

allennlp_models/vision/dataset_readers/flickr30k.py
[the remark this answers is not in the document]
jacob-morrison: Sounds good!

allennlp_models/vision/dataset_readers/flickr30k.py
```diff
@@ -0,0 +1,469 @@
...
+            coords = [TensorField(coordinates_list[image_index])]
+            masks = [
+                ArrayField(
+                    features_list[image_index].new_ones(
+                        (features_list[image_index].shape[0],), dtype=torch.bool
+                    ),
+                    padding_value=False,
+                    dtype=torch.bool,
+                )
+            ]
```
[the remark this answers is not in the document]
jacob-morrison: I think so too, but I can't remember what the reasoning was? All the other image dataset readers explicitly create the masks, so I assume I should do that here as well.

allennlp_models/vision/models/vilbert_image_retrieval.py
dirkgr: Yes, you'd have to have two different metrics, because the metric forces you to specify `k` in the constructor.

allennlp_models/vision/dataset_readers/flickr30k.py
```diff
@@ -0,0 +1,469 @@
...
+
+    def __init__(
+        self,
+        image_dir: Union[str, PathLike],
+        *,
+        image_loader: Optional[ImageLoader] = None,
+        image_featurizer: Optional[Lazy[GridEmbedder]] = None,
+        region_detector: Optional[Lazy[RegionDetector]] = None,
+        feature_cache_dir: Optional[Union[str, PathLike]] = None,
+        data_dir: Optional[Union[str, PathLike]] = None,
```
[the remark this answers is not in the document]
jacob-morrison: They're actually slightly different. The `file_path` is a link to a text file that contains the image IDs of all the images in that split, and the `data_dir` contains all of the caption files that are associated with each image. So we read the file at `file_path` to collect a list of image names, and then we read the relevant caption files in `data_dir`. I'll switch `data_dir` to `cached_path`, so we can use a URL.

allennlp_models/vision/dataset_readers/flickr30k.py
```diff
@@ -0,0 +1,469 @@
...
+            tokenizer=tokenizer,
+            token_indexers=token_indexers,
+            cuda_device=cuda_device,
+            max_instances=max_instances,
+            image_processing_batch_size=image_processing_batch_size,
+            write_to_cache=write_to_cache,
+            manual_distributed_sharding=False,
+            manual_multiprocess_sharding=False,
+        )
+        self.data_dir = data_dir
```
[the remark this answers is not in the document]
jacob-morrison: Sounds good, I'll test that now.

allennlp_models/vision/dataset_readers/flickr30k.py
```diff
@@ -88,6 +90,12 @@ class Flickr30kReader(VisionReader):
         the captions and metadata for each task instance.
     tokenizer: `Tokenizer`, optional
     token_indexers: `Dict[str, TokenIndexer]`
+    featurize_captions: `bool`, optional
+        If we should featurize captions while calculating hard negatives, or use placeholder features.
+    is_evaluation: `bool`, optional
+        If the reader should return instances for evaluation or training.
+    n: int, optional
+        The number of potential hard negatives to consider.
```
dirkgr, line 98: Call it `num_hard_negatives` then?

dirkgr: Has that always been configurable? I thought we were always using one hard negative?

allennlp_models/vision/dataset_readers/flickr30k.py
```diff
@@ -136,13 +144,55 @@ def __init__(
...
     def _read(self, file_path: str):
         file_path = cached_path(file_path, extract_archive=True)
         files_in_split = set()
+        i = 0
         with open(file_path, "r") as f:
             for line in f:
+                if self.max_instances is not None and i >= self.max_instances:
+                    break
                 files_in_split.add(line.rstrip("\n"))
+                i += 5
```
dirkgr: You can instead write something like this:
```Python
for i, line in enumerate(f):
    ...
```
... and then compare to `I*5`. Then you don't need to maintain that counter variable.

allennlp_models/vision/dataset_readers/flickr30k.py
```diff
@@ -151,64 +201,51 @@ def _read(self, file_path: str):
...
+            processed_images = self._process_image_paths(
+                self.images[filename] for filename in tqdm.tqdm(filenames, desc="Processing images")
+            )
+        except KeyError as e:
+            missing_id = e.args[0]
+            raise KeyError(
+                missing_id,
+                f"We could not find an image with the id {missing_id}. "
+                "Because of the size of the image datasets, we don't download them automatically. "
+                "Please go to https://visualqa.org/download.html, download the datasets you need, "
```
dirkgr, line 215: This is not the correct URL, is it?

tests/vision/dataset_readers/flickr30k_test.py
```diff
@@ -94,14 +94,14 @@ def test_evaluation_read(self):
...
-            "without",
+            "with",
             "brown",
             "hair",
             "sits",
             "on",
             "edge",
             "of",
-            "cement",
+            "concrete",
```
dirkgr: What are these changes for?

dirkgr (changes requested):
Only small changes now

allennlp_models/vision/dataset_readers/flickr30k.py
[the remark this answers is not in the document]
dirkgr: Let's leave it in the interest of time. Maybe revisit later.

tests/vision/dataset_readers/flickr30k_test.py
jacob-morrison: Some seeds would select a random "wrong" caption, and then accidentally pick the same caption, because the 3 caption files were the same. Then the test would fail because it would compare two captions that should be different, but weren't. I made a few small tweaks to fix it.

allennlp_models/vision/dataset_readers/flickr30k.py
jacob-morrison: It's not, thanks for catching that

allennlp_models/vision/dataset_readers/flickr30k.py
jacob-morrison: Thanks!

allennlp_models/vision/dataset_readers/flickr30k.py
jacob-morrison: This is a different variable, that I guess doesn't need to be configurable. We first find the `n` (default = `100`) closest images based on the L2 distance between their averaged features, then out of those `n`, we do the actual calculations with the caption vectors. I'm fine with removing it or renaming it, what do you think?

tests/vision/dataset_readers/flickr30k_test.py
```diff
@@ -8,11 +8,13 @@
 from allennlp.modules.vision.grid_embedder import NullGridEmbedder
 from allennlp.modules.vision.region_detector import RandomRegionDetector
 
+import random
 from tests import FIXTURES_ROOT
 
 
 class TestFlickr30kReader(AllenNlpTestCase):
     def setup_method(self):
+        random.seed(12345)
```
dirkgr, line 17: What's the deal with this? Why does that make the tests less random?

jacob-morrison: It doesn't, I was testing something but it didn't work. The test is passing locally, so it's hard to debug what's going wrong.

PR #287, opened 2021-06-28
Add label smoothing to CopyNet

JohnGiorgi:
This PR adds label smoothing to [`CopyNetSeq2Seq`](https://github.com/allenai/allennlp-models/blob/main/allennlp_models/generation/models/copynet_seq2seq.py). As discussed in https://github.com/allenai/allennlp/issues/5276, label smoothing is added to the generation scores only. It is mostly a re-working of the existing label smoothing code in [`sequence_cross_entropy_with_logits`](https://github.com/allenai/allennlp/blob/cf113d705b9054d329c67cf9bb29cbc3f191015d/allennlp/nn/util.py#L821-L836).

As a sanity check, I ran the code with my own model. A model with a small `label_smoothing` value reaches similar performance as a model with `label_smoothing == 0.0`. As for additional unit tests, I think a modification of [`test_get_ll_contrib`](https://github.com/allenai/allennlp-models/blob/e47da9911155c7d65bf0dfba51aed048d367e5ea/tests/generation/models/copynet_test.py#L67) might make the most sense.

### DIFF
--- a/allennlp_models/generation/models/copynet_seq2seq.py
+++ b/allennlp_models/generation/models/copynet_seq2seq.py
@@ -458,7 +458,7 @@ class CopyNetSeq2Seq(Model):
                 -1, target_tokens.unsqueeze(1), 1.0 - self._label_smoothing
             )
             smoothed_targets = one_hot_targets + smoothing_value
-            generation_log_probs = -log_probs * smoothed_targets
+            generation_log_probs = log_probs * smoothed_targets
             # shape: (batch_size, 1)
             generation_log_probs = generation_log_probs.sum(-1, keepdim=True)
         else:
--- a/allennlp_models/generation/models/copynet_seq2seq.py
+++ b/allennlp_models/generation/models/copynet_seq2seq.py
@@ -455,7 +455,7 @@ class CopyNetSeq2Seq(Model):
             # Create the smoothed targets to be multiplied with `log_probs`
             # shape: (batch_size, target_vocab_size + source_sequence_length)
             smoothed_targets = torch.full_like(log_probs, smoothing_value).scatter_(
-                -1, target_tokens.unsqueeze(1), 1.0 - self._label_smoothing + smoothing_value
+                1, target_tokens.unsqueeze(1), 1.0 - self._label_smoothing + smoothing_value
             )
             generation_log_probs = log_probs * smoothed_targets
             # shape: (batch_size, 1)
--- a/allennlp_models/generation/models/copynet_seq2seq.py
+++ b/allennlp_models/generation/models/copynet_seq2seq.py
@@ -452,12 +452,11 @@ class CopyNetSeq2Seq(Model):
         # Now we get the generation score for the gold target token.
         if self._label_smoothing is not None and self._label_smoothing > 0.0:
             smoothing_value = self._label_smoothing / target_size
-            # Fill all the correct indices with 1 - smoothing value.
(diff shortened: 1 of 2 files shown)


allennlp_models/generation/models/copynet_seq2seq.py
```diff
@@ -443,8 +450,24 @@ def _get_ll_contrib(
...
-        # shape: (batch_size, 1)
-        generation_log_probs = log_probs.gather(1, target_tokens.unsqueeze(1)) + log_gen_mask
+        if self._label_smoothing is not None and self._label_smoothing > 0.0:
+            smoothing_value = self._label_smoothing / target_size
+            # Fill all the correct indices with 1 - smoothing value.
+            # shape: (batch_size, target_vocab_size + source_sequence_length)
+            one_hot_targets = torch.zeros_like(log_probs).scatter_(
+                -1, target_tokens.unsqueeze(1), 1.0 - self._label_smoothing
+            )
+            smoothed_targets = one_hot_targets + smoothing_value
```
dirkgr, line 460: It seems like it would be faster to start with `torch.full_like(log_probs, smoothing_value)` and go from there? I did not measure it though.

`one_hot_targets` isn't used anywhere else, is it? It's not even properly "one hot" like this. Maybe it should be called `one_warm_targets` 🤣 .

dirkgr (approved):
Looks good, just a small comment.

dirkgr:
Oh, I see it's also missing a changelog entry. This definitely has enough magnitude to warrant one.

allennlp_models/generation/models/copynet_seq2seq.py
JohnGiorgi: > It seems like it would be faster to start with `torch.full_like(log_probs, smoothing_value)` and go from there? I did not measure it though.

Something like 

```python
one_hot_targets = torch.full_like(log_probs, smoothing_value).scatter_(
    -1, target_tokens.unsqueeze(1), 1.0
)
```

I can do a little test to see if that's faster!

> `one_hot_targets` isn't used anywhere else, is it? It's not even properly "one hot" like this. Maybe it should be called `one_warm_targets` 🤣 .

Good point! I am stealing the variable `one_hot_targets` right from [here](https://github.com/allenai/allennlp/blob/cf113d705b9054d329c67cf9bb29cbc3f191015d/allennlp/nn/util.py#L821-L836). Another solution is just to use `smoothed_targets`:

```python
smoothed_targets = torch.zeros_like(log_probs).scatter_(
    -1, target_tokens.unsqueeze(1), 1.0 - self._label_smoothing
)
smoothed_targets = smoothed_targets + smoothing_value
```

or even 

```python
smoothed_targets = torch.zeros_like(log_probs).scatter_(
    -1, target_tokens.unsqueeze(1), 1.0 - self._label_smoothing
) + smoothing_value
```

but that might be putting too much logic into one line.

JohnGiorgi:
> Oh, I see it's also missing a changelog entry. This definitely has enough magnitude to warrant one.

Done!

allennlp_models/generation/models/copynet_seq2seq.py
JohnGiorgi: Might be worth updating https://github.com/allenai/allennlp/blob/cf113d705b9054d329c67cf9bb29cbc3f191015d/allennlp/nn/util.py#L825-L828 to use this micro-optimization

dirkgr:
Thanks! I'll make the switch in util.py.

PR #288, opened 2021-07-19
Pass vocab to beam search

JohnGiorgi:
This is a small PR that should be considered alongside https://github.com/allenai/allennlp/pull/5321. It just passes `vocab` to all the `beam_search.construct` calls in the `generation` models.


epwalsh (approved):
LGTM!

PR #292, opened 2021-07-27
Bugfix: initializing all tensors and parameters of the `ConditionalRandomField` on the proper device

oroszgy:
Fixes https://github.com/allenai/allennlp/issues/2884

For details see: https://github.com/allenai/allennlp/pull/5335

### DIFF
--- a/allennlp_models/tagging/models/crf_tagger.py
+++ b/allennlp_models/tagging/models/crf_tagger.py
@@ -297,6 +297,12 @@ class CrfTagger(Model):
         return output_dict
 
     @overrides
+    def cuda(self, device=None):
+        super().cuda(device)
+        self.crf.cuda(device)
+        return self
+
+    @overrides
     def get_metrics(self, reset: bool = False) -> Dict[str, float]:
         metrics_to_return = {
             metric_name: metric.get_metric(reset) for metric_name, metric in self.metrics.items()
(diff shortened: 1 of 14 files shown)


allennlp_models/tagging/models/crf_tagger.py
```diff
@@ -296,6 +296,12 @@ def decode_top_k_tags(top_k_tags):
 
         return output_dict
 
+    @overrides
+    def cuda(self, device=None):
+        super().cuda(device)
+        self.crf.cuda(device)
+        return self
```
epwalsh, line 303: Is this really necessary? `self.crf` is a `nn.Module`, so I would think this already works fine without overriding this method.

dirkgr:
What's the deal with this PR? The title makes it seem important.

allennlp_models/tagging/models/crf_tagger.py
oroszgy: @epwalsh you were true, just checked it. We can safely remove this change. The bug is only in the `conditional_random_field.py`

oroszgy:
@dirkgr I think you can close this PR, this should solve alone the problem: https://github.com/allenai/allennlp/pull/5335

allennlp_models/tagging/models/crf_tagger.py
oroszgy: Fixed: https://github.com/allenai/allennlp/pull/5335/commits/84d4e20282ef05a16762cf322683e5b3512db8ab

PR #293, opened 2021-07-29
Updating vision reader to also produce class probs and labels

AkshitaB:
`RegionDetectorOutput` also contains class probs and labels, depending on the region detector selected. The dataset readers inheriting from `VisionReader` should be able to use this information when available.

### DIFF
--- a/allennlp_models/vision/dataset_readers/vision_reader.py
+++ b/allennlp_models/vision/dataset_readers/vision_reader.py
@@ -2,6 +2,7 @@ import glob
 import logging
 from os import PathLike
 from typing import (
+    Any,
     Dict,
     List,
     Union,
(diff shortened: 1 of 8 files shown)


allennlp_models/vision/dataset_readers/vision_reader.py
```diff
@@ -234,9 +239,37 @@ def _coordinates_cache(self) -> MutableMapping[str, Tensor]:
...
+            if self.class_labels_cache_dir is None:
+                self._class_labels_cache_instance = {}
+            else:
+                os.makedirs(self.feature_cache_dir, exist_ok=True)  # type: ignore
+                self._class_labels_cache_instance = TensorCache(
+                    os.path.join(self.feature_cache_dir, "class_labels"),  # type: ignore
+                    read_only=not self.write_to_cache,
+                )
+
+        return self._class_labels_cache_instance
```
dirkgr, line 268: We have this code in here four times. Any chance we can simplify, maybe with the help of https://docs.python.org/3/library/functools.html#functools.cached_property?

dirkgr (approved):
Seems fine, but it would be nice if we could get rid of that duplicated code

allennlp_models/vision/dataset_readers/vision_reader.py
AkshitaB: `cached_property` isn't available in `python3.7`, I think. I've updated the code to reduce duplicate code, however.

PR #296, opened 2021-08-18
Fix a bug in the last decoding step of training CopyNet

GregoryZeng:
Closes https://github.com/allenai/allennlp/issues/5354

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Fixed tests for Spacy versions greater than 3.1
+- Fixed the last step decoding when training CopyNet
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


epwalsh (commented):
Thanks @GregoryZeng! This is great. I think the comment on [lines 495-496](https://github.com/allenai/allennlp-models/blob/31649f5776522ae60661499c595f73e1c019d72c/allennlp_models/generation/models/copynet_seq2seq.py#L495-L496) is still confusing:
```python
        # The last input from the target is either padding or the end symbol.
        # Either way, we don't have to process it.
        num_decoding_steps = target_sequence_length - 1
```
I think it would be more clear if it said something like:
```python
        # We have a decoding step for every target token except the "START" token.
        num_decoding_steps = target_sequence_length - 1
```
Do you agree? If so, could you update that in this PR as well?

GregoryZeng:
Sure, that's definitely clearer.

epwalsh (approved):
Looks great! Thanks @GregoryZeng!

PR #297, opened 2021-08-20
IMDB Model

dirkgr:
I will need this for DSPT.

This depends on https://github.com/allenai/allennlp/pull/5372 being merged.

### DIFF
--- a/allennlp_models/classification/models/transformer_classification_tt.py
+++ b/allennlp_models/classification/models/transformer_classification_tt.py
@@ -32,7 +32,7 @@ class TransformerClassificationTransformerToolkit(Model):
         num_labels: Optional[int] = None,
         label_namespace: str = "labels",
         override_weights_file: Optional[str] = None,
-        **kwargs
+        **kwargs,
     ) -> None:
         super().__init__(vocab, **kwargs)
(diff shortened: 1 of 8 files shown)


allennlp_models/classification/models/transformer_classification_tt.py
```diff
@@ -0,0 +1,103 @@
...
+from allennlp.data import Vocabulary
+from allennlp.models import Model
+from allennlp.modules.transformer import TransformerEmbeddings, TransformerStack, TransformerPooler
+from torch.nn import Dropout
+
+logger = logging.getLogger(__name__)
+
+
+@Model.register("transformer_classification_tt")
+class TransformerClassificationTransformerToolkit(Model):
```
AkshitaB, line 14: That's a verbose name! :) Is there a reason to name models that use the transformer toolkit thus?

dirkgr: I did it this way for the "MC" models, and I wanted to stay consistent. But "MC" is much shorter than "Classification", so now it looks silly.

We could rename all the Transformer Toolkit models to just "TT"? Then we also don't have to say "Transformer" twice.

PR #299, opened 2021-08-31
MaskedLM

dirkgr:
This is the other half of the changes in https://github.com/allenai/allennlp/pull/5391.


dirkgr:
I don't think I need this anymore.

PR #306, opened 2021-10-25
Fix the `IndexError` when `CNNDailyMailDatasetReader` reads test data.

xinzhel:
Problem: `IndexError` happens when the first line of a text file only contains `(CNN)`. This happens for reading test data of CNN news.

Example: You can check this file in the extracted `cnn_stories` directory: `12078b09d95c01cedb06da7fc63faab540432dee.story`.
This file would be read for test data (corresponding to this URL in all_text.txt`: http://web.archive.org/web/20150617021105id_/http://www.cnn.com/2015/04/16/opinions/medical-marijuana-revolution-sanjay-gupta/)

Note: the provided [training config](https://github.com/allenai/allennlp-models/blob/main/training_config/generation/bart_cnn_dm.jsonnet) does not evaluate on test data. So this error could be ignored in this case. 

Solution: remove "(CNN)" before checking the empty string


epwalsh (approved):
LGTM! Thanks!

PR #309, opened 2021-11-22
Add option to use scheduled sampling in CopyNet

JohnGiorgi:
This PR adds the ability to use scheduled sampling in `CopyNetSeq2Seq` by supplying an argument for `scheduled_sampling_ratio` that's greater than zero. It is essentially a copy/paste from `SimpleSeq2Seq`.

This helps reduce the differences in the `SimpleSeq2Seq` and `CopyNetSeq2Seq` model arguments. It is also backwards compatible with a default value of 0 (no scheduled sampling i.e. teacher forcing).


epwalsh (changes requested):
Thanks @JohnGiorgi! I think this is a good addition. I just have a couple of suggestions:
- I think we should have a test
- I think the default for `scheduled_sampling_ratio` should be `None` so that when it is `None` we don't have to call `torch.rand()`. That way there is no performance penalty for this feature.

JohnGiorgi:
@epwalsh Awesome, thanks for the feedback. 

- Updated the `if` statement so that if `scheduled_sampling_ratio` is falsey, `torch.rand` is not called. You could set the default to `None`, but this makes the default value for this parameter different than `simple_seq2seq`.
- Added a simple test that the model can train with `0 < scheduled_sampling_ratio < 1`. I wasn't really sure how to write a more significant test due to the randomness. Let me know if you have a different test in mind!

I could also update `simple_seq2seq` so that `torch.rand` is not called when `scheduled_sampling_ratio` is falsey, to save a tiny bit of performance there!

allennlp_models/generation/models/copynet_seq2seq.py
```diff
@@ -514,13 +524,33 @@ def _forward_loss(
...
-            copied = ((input_choices == self._oov_index) & (target_to_source.sum(-1) > 0)).long()
+            if (
+                self.training
+                and self._scheduled_sampling_ratio
+                and torch.rand(1).item() < self._scheduled_sampling_ratio
+            ):
+                # Use gold tokens at test time and at a rate of 1 - _scheduled_sampling_ratio
+                # during training.
+                # shape: (batch_size,)
+                input_choices = last_predictions
```
JohnGiorgi, line 543: Ah, I realized this implementation doesn't work, because `last_predictions` is never updated. I would have had to take the index of the token with the highest probability for this timestep under the model. Something like:

```python
last_predictions = torch.max(torch.cat((copy_scores, generation_scores), -1), -1)
```

@epwalsh does this make sense?

tests/generation/models/copynet_test.py
```diff
@@ -44,6 +44,13 @@ def test_model_can_train_with_amp(self):
             overrides="{'trainer.use_amp':true,'trainer.cuda_device':0}",
         )
 
+    def test_model_can_train_with_scheduled_sampling_ratio(self):
+        train_model_from_file(
+            self.param_file,
+            self.TEST_DIR,
+            overrides="{'model.scheduled_sampling_ratio':0.5}",
+        )
```
epwalsh, line 52: > Added a simple test that the model can train with 0 < scheduled_sampling_ratio < 1. I wasn't really sure how to write a more significant test due to the randomness. Let me know if you have a different test in mind!

I think this is totally fine for a test. I just want to have coverage of the new lines of code to ensure that the model trains without error.

allennlp_models/generation/models/copynet_seq2seq.py
JohnGiorgi: Gotcha. Could I just return [`log_probs`](https://github.com/allenai/allennlp-models/blob/86444374c487f1d6991cd83b7a3da80b026effc2/allennlp_models/generation/models/copynet_seq2seq.py#L430) from `_get_ll_contrib()`? Its computed like: `log_probs = util.masked_log_softmax(all_scores, mask)`.

epwalsh: Yes, good point.

JohnGiorgi: Awesome, just pushed that change.

tests/generation/models/simple_seq2seq_test.py
```diff
@@ -46,6 +46,13 @@ def test_model_can_train_with_amp(self):
             overrides="{'trainer.use_amp':true,'trainer.cuda_device':0}",
         )
 
+    def test_model_can_train_with_scheduled_sampling_ratio(self):
+        train_model_from_file(
+            self.param_file,
+            self.TEST_DIR,
+            overrides="{'model.scheduled_sampling_ratio':0.5}",
+        )
+
```
JohnGiorgi, line 55: @epwalsh Added the sample test for scheduled sampling to `simple_seq2seq`.

allennlp_models/generation/models/simple_seq2seq.py
```diff
@@ -367,7 +367,11 @@ def _forward_loop(
         step_logits: List[torch.Tensor] = []
         step_predictions: List[torch.Tensor] = []
         for timestep in range(num_decoding_steps):
-            if self.training and torch.rand(1).item() < self._scheduled_sampling_ratio:
+            if (
+                self.training
+                and self._scheduled_sampling_ratio > 0.0
+                and torch.rand(1).item() < self._scheduled_sampling_ratio
+            ):
```
JohnGiorgi, line 374: @epwalsh Added a similar condition to `simple_seq2seq` to avoid the call to `torch.rand` when `_scheduled_sampling_ratio` is `0.0`.

epwalsh (commented):
This LGTM! Can you just update the CHANGELOG? Then I think this is good to go.

JohnGiorgi:
Cool! Changelog updated 👍

epwalsh (approved):

PR #313, opened 2021-12-01
Removes stuff that now lives in the tango repo

dirkgr:
The title of the PR is not strictly true. There is no equivalent of the multiple choice training configs. But it would be a lot of work to put them back in, and we decided that seq2seq models are more important.


epwalsh (approved):
LGTM

PR #322, opened 2022-02-23
Missing perl error message

dirkgr:
Fixes https://github.com/allenai/allennlp/issues/5573.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -533,7 +533,7 @@ jobs:
     - name: Patch AllenNLP dependency
       if: github.event_name == 'push' || github.event_name == 'pull_request'
       run: |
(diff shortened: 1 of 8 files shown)


dirkgr:
There is now nothing left of this PR, but thanks to squash merging, GitHub can't figure this out automatically.

Denbergvanthijs:
Thank you very much! Just a quick suggestion: wouldn't `from shutil import which` be a cleaner option instead of a try-except? [It looks to be multiplatform ](https://docs.python.org/3/library/shutil.html) and available since Python 3.3. When the executable is not found, it returns None.

dirkgr:
While unlikely, that introduces a race condition. What if someone moves the perl executable after your call to `which` and before the call to `subprocess.run`?

Denbergvanthijs:
That would indeed be a possibility, thank you for your answer!

PR #323, opened 2022-02-24
Update CI to install allennlp with optional deps for tests

AkshitaB:


.github/workflows/ci.yml
```diff
@@ -166,6 +166,7 @@ jobs:
     - name: Install requirements
       run: |
         . .venv/bin/activate
+        pip install "allennlp[all] @ git+https://github.com/allenai/allennlp.git@main"
```
epwalsh, line 169: I think we want to change this in `requirements.txt` instead of here.

dirkgr: We don't want the release of -models to depend on the main branch of allennlp though.

epwalsh: We already handle that with the `ALLENNLP_VERSION_OVERRIDES` env variable

dirkgr: For development, I always have to change that line in `requirements.txt` to just say `allennlp`, so that `pip install -e ./allennlp -e ./allennlp-models` works.

epwalsh: https://github.com/allenai/allennlp-models#from-source

AkshitaB: I thought to do that too at first, but the models repo doesn't need checklist. Only the tests do.

epwalsh: Oh, then you could set `ALLENNLP_VERSION_OVERRIDE` to that

epwalsh (approved):
LGTM!

PR #324, opened 2022-02-28
Add option to specify the image extensions to be loaded/discovered

shunk031:
Thanks for the great work on this repository. I am helped every day by allennlp and allennlp-models. I think this PR is a small contribution, but I'm glad I could help this awesome library.

Anyway, this PR adds an option regarding image extensions to `__init__` of `VisionReader`. 

Currently, the VisionReader only discovers images with `jpg` and `png` ([ref.](https://github.com/allenai/allennlp-models/blob/v2.9.0/allennlp_models/vision/dataset_readers/vision_reader.py#L188)).
The option allows us to discover and load images with extensions not limited to jpg and png (e.g, `jpeg` or `bmp` etc.).

If you find this PR useful, I would appreciate it if you could merge it. Thanks!


AkshitaB (approved):
Hi @shunk031 Thanks for the PR. Can you also update the changelog?

shunk031:
@AkshitaB Thank you for checking my PR. I updated the CHANGELOG. Cloud you check the file?

AkshitaB (changes requested):
@shunk031 The `allennlp-models` library goes along with `allennlp`. Right now, this code will discover other extensions, but the [image_loader](https://github.com/allenai/allennlp/blob/main/allennlp/data/image_loader.py) will not be able to read those files (`torchvision.io.read_image` only reads png/jpg). 

My suggestion would be to have a fixed list of acceptable extensions (instead of passing them as an argument), and have a corresponding PR in `allennlp` that reads those extensions correctly.

shunk031:
The primary motivation of my PR is to allow `jpeg` file to discover from the VisionReader. This motivation can be satisfied by simply adding `jpeg` to the set of extensions. Can I just add the jpeg extension of the set of the extensions? If so, do I need to close this PR and create a new one to add the extension? Thanks!

AkshitaB:
@shunk031 Yes, you can add the jpeg extension. There's no need to create a separate PR, but do add a test case to make sure that all 3 extensions are detected.

shunk031:
@AkshitaB It is almost complete, including the test code.  I will add 224x224 sized png, jpg, and jpeg files like the following image I created to this repository for testing, is that ok?

png_example

AkshitaB:
@shunk031 You can also use existing test_fixture images: https://github.com/allenai/allennlp-models/tree/main/test_fixtures/vision/images

AkshitaB:
@shunk031 You can also use existing test_fixture images: https://github.com/allenai/allennlp-models/tree/main/test_fixtures/vision/images

AkshitaB (approved):

PR #341, opened 2022-06-22
Implementation of Weighted CRF Tagger (handling unbalanced datasets)

eraldoluis:
Closes [allennlp issue #4619](https://github.com/allenai/allennlp/issues/4619).

Depends on [allennlp PR #5676](https://github.com/allenai/allennlp/pull/5676)

Changes proposed in this pull request:

- I implemented and [experimentally compared](https://eraldoluis.github.io/2022/05/10/weighted-crf.html) three sample weighting strategies for CrfTagger.
- I added two parameters to CrfTagger: `label_weights` and `weight_strategy`.
- The parameter `label_weights` is a `Dict[str, float]` with a mapping {label : weight} to be used in the loss function in order to give different weights for each token depending on its label.
- The parameter `weight_strategy` can be: `None` `'emission'`, `'emission_transition'` or `'lannoy'`.
- If `label_weights` is given and `weight_strategy` is `None` or `'emission'`, then the emission score of each tag is multiplied by the corresponding weight (as given by `label_weights`).
- If `emission_transition`, both emission and transition scores of each tag are multiplied by the corresponding weight.
- If `weight_strategy` is `'lannoy'`, then we use the strategy proposed by [Lannoy et al. (2019)](https://perso.uclouvain.be/michel.verleysen/papers/ieeetbe12gdl.pdf).
- An experimental comparison among these three strategies and a brief discussion of their differences [here](https://eraldoluis.github.io/2022/05/10/weighted-crf.html).
- Tests were created to cover the new feature.


CHANGELOG.md
```diff
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- Changed the token-based verbose metric (when `verbose_metrics` is `True` and `calculate_span_f1` is `False`) to be `FBetaVerboseMeasure` instead of `FBetaMeasure`.
```
epwalsh, line 12:
```suggestion
- Changed the token-based verbose metric in the `CrfTagger` model (when `verbose_metrics` is `True` and `calculate_span_f1` is `False`) to be `FBetaVerboseMeasure` instead of `FBetaMeasure`.
```

epwalsh (approved):
LGTM!

PR #344, opened 2022-07-09
Remove duplicate TextField

deldelmax:
Seems like the tokens get wrapped in a TextField twice. Interesting that the model's code doesn't break.


dirkgr:
Looks good. Can you make a changelog entry so we can merge this?

deldelmax:
> Looks good. Can you make a changelog entry so we can merge this?

Yes, updated.

dirkgr (approved):

PR #346, opened 2022-11-23
Remove upper bounds for dependencies in `requirements.txt`

aphedges:
This is a follow-up to <https://github.com/allenai/allennlp/commit/80fb6061e568cb9d6ab5d45b661e86eb61b92c82> to make installation after sunset easier.

---

This PR is supposed to achieve the same goal as <https://github.com/allenai/allennlp/pull/5733>. @epwalsh, do you want to review this one as well?


aphedges:
I have manually added the constraints from <https://github.com/allenai/allennlp/commit/80fb6061e568cb9d6ab5d45b661e86eb61b92c82> so the correct transitive dependencies will be installed during testing. I hope it works.

aphedges:
The Lint failure was caused by a `flake8` update an hour ago. Pinning the package version fixes it.

The CPU Tests are failing with the same test and error message in #345, and they are failing with the same test but a different error on `main`, so it seems like a completely separate problem from this PR.

dirkgr:
Yes, I'm trying to get the tests to run before I look at these other PRs. Thanks for contributing!

epwalsh (approved):
LGTM

PR #347, opened 2022-11-23
Fix test that only fails locally on linux

dirkgr:


dirkgr:
This is merged now, but since we squash instead or merging properly, git can't figure this out.

